PR #3, opened 2017-04-22
Turns createApps and command center setup synchronous

ronkorving:
These used to be asynchronous, but since it's during the startup phase only, synchronous doesn't hurt. More importantly, this makes `webpack.config.js` a lot easier to implement, since you don't need to mess about with promises and asynchronous flow during configuration.

After making those changes, I've also made `createApps` unnecessary when calling `mage.getClientConfig(appName)`. Now, that method will automatically create the app if it wasn't created yet. On top of that, `createApps()` will ignore any apps already created, so it's perfectly safe to create one app first, and all other apps later for example. This change further simplified the Webpack config.

**BC breakage**: Nobody was using this, but regardless, this PR removes configuration entry `apps: appName: appConfig`.

**BC breakage**: `mage.core.app.createApps(cb)` no longer takes a callback, as it's now synchronous. Normally this method is called by mage itself, but build configurations (ie: Webpack) have called it to do a partial mage setup. That is however no longer needed.

**BC breakage**: This removes the `baseUrl` argument from `mage.getClientConfig(appName, baseUrl)` and pulls the information out of httpServer (which gets it from configuration). That should make its usage more intuitive. Up until now, all applications that I'm aware of passed empty string anyway, so this change should not have a negative impact.

**BC breakage**: Configuration `server.clientHost.expose` no longer supports object notation, and now only allows strings. As far as I'm aware, no existing projects use this notation anyway, but this really keeps things simpler. Non-strings will throw. Tests have been updated accordingly, and coverage is good.

Testing: Changes are all fully covered. I've also successfully tested this on a project.

Typical impact on `webpack.config.js`:

> Before: webpack.config.js

```js
'use strict';

const webpack = require('webpack');

const baseUrl = '';

let config;
let devMode;

function setupApp() {
	return new Promise((resolve, reject) => {
		const mage = require('./lib');

		mage.core.app.createApps((error) => {
			if (error) {
				return reject(error);
			}

			config = mage.getClientConfig('api', baseUrl);
			devMode = mage.isDevelopmentMode();

			return resolve();
		});
	});
}


module.exports = setupApp().then(() => {
	return {
                // ... config properties
		plugins: [
			new webpack.DefinePlugin({
				MAGE_CONFIG: JSON.stringify(config),
				DEBUG: JSON.stringify(devMode)
			})
		]
	};
});
```

> After: webpack.config.js

```js
'use strict';

const webpack = require('webpack');
const mage = require('./lib');

module.exports = {
        // ... config properties
	plugins: [
		new webpack.DefinePlugin({
			MAGE_CONFIG: JSON.stringify(mage.getClientConfig('api')),
			DEBUG: JSON.stringify(mage.isDevelopmentMode())
		})
	]
};
```

### DIFF
--- a/lib/app/Readme.md
+++ b/lib/app/Readme.md
@@ -8,12 +8,12 @@ In your configuration, you can enable an app using something like this:
 
 ```yaml
 apps:
-    # Replace appName with the name of the app you wish to enable
-    appName:
-        responseCache: 180
-        disabled: false
+  # Replace appName with the name of the app you wish to enable
+  appName:
+    responseCache: 180
+    disabled: false
 ```
 
-* `appName` (mandatory) This is the name of the app you wish to enable.
-* `responseCache` (optional) Number of seconds usercommand responses should be cached for. (default: 3 mins)
-* `disabled` (optional) This will determine if the app gets exposed at all. (default: true)
+* `appName` (string; mandatory) This is the name of the app you wish to enable.
+* `responseCache` (number; optional) Number of seconds usercommand responses should be cached for. (default: 3 mins)
+* `disabled` (boolean; optional) This will determine if the app gets exposed at all. (default: true)
--- a/lib/app/index.js
+++ b/lib/app/index.js
@@ -1,53 +1,180 @@
-var async = require('async');
+'use strict';
 
-var mage = require('../mage');
-var logger = mage.core.logger;
+/** @module app */
+
+const memoize = require('memoizee');
 
 
 // app registration
 
-var appMap = {};
-var appList = [];
+const appMap = {};
+const appList = [];
 
 
 /**
- * This function creates instances of all configured apps and registers them into appMap and appList.
+ * Returns the configuration for all apps (overridden during unit tests)
+ *
+ * @returns {Object}
  */
+/* istanbul ignore next */
+exports.getAppsConfig = memoize(function () {
+	return require('lib/mage').core.config.get(['apps']);
+});
 
-exports.createApps = function (cb) {
-	var appsConfig = mage.core.config.get(['apps']);
+/**
+ * Returns a logger (overridden during unit tests)
+ *
+ * @returns {Object}
+ */
+/* istanbul ignore next */
+exports.getLogger = memoize(function () {
+	return require('lib/mage').core.logger;
+});
 
-	var appNames = Object.keys(appsConfig);
+/**
+ * Creates and returns a CommandCenter instance for the given app (overridden during unit tests)
+ *
+ * @returns {Object}
+ */
+/* istanbul ignore next */
+exports.createCommandCenter = function (app) {
+	const CommandCenter = require('lib/mage').core.cmd.CommandCenter;
+	return new CommandCenter(app);
+};
 
-	async.eachLimit(appNames, 5, function (appName, callback) {
-		var appConfig = appsConfig[appName];
 
-		if (!appConfig || appConfig.disabled) {
-			logger.debug('Not creating app:', appName, appConfig ? '(disabled)' : '(no configuration found)');
-			return setImmediate(callback);
+/**
+ * @classdesc App
+ */
+class App {
+	/**
+	 * @constructor
+	 * @param {string} name     The name of the application
+	 * @param {Object} config   The application configuration
+	 */
+	constructor(name, config) {
+		this.name = name;
+		this.config = config || {};
+		this.commandCenter = exports.createCommandCenter(this);
+	}
+
+	/**
+	 * Sets up the command center for this app
+	 */
+	setup() {
+		this.commandCenter.setup();
+	}
+}
+
+
+/**
+ * Creates an instance of a configured app and registers it into appMap and appList
+ *
+ * @param {string} name  The name of the application
+ * @returns {Object}     The created application
+ * @throws {Error}       If the app by the given name is not configured to exist or has already been created
+ */
+exports.createApp = function (name) {
+	const logger = exports.getLogger();
+	logger.debug('Creating app:', name);
+
+	const appsConfig = exports.getAppsConfig();
+	if (!appsConfig.hasOwnProperty(name)) {
+		throw new Error(`App "${name}" has not been configured`);
+	}
+
+	if (appMap[name]) {
+		throw new Error(`App "${name}" has already been created`);
+	}
+
+	const app = new App(name, appsConfig[name]);
+	app.setup();
+
+	appMap[name] = app;
+	appList.push(app);
+
+	return app;
+};
+
+
+/**
+ * Creates instances of all configured apps and registers them into appMap and appList.
+ */
+exports.createApps = function () {
+	const logger = exports.getLogger();
+	const appsConfig = exports.getAppsConfig();
+
+	const appNames = Object.keys(appsConfig);
+	for (const appName of appNames) {
+		const appConfig = appsConfig[appName];
+
+		if (appMap[appName]) {
+			logger.debug(`Not creating app "${appName}" (already created)`);
+			continue;
 		}
 
-		logger.debug('Creating app:', appName);
+		if (!appConfig) {
+			logger.debug(`Not creating app "${appName}" (no configuration found)`);
+			continue;
+		}
 
-		var app = {
-			name: appName,
-			config: appConfig,
-			commandCenter: null
-		};
+		if (appConfig.disabled) {
+			logger.debug(`Not creating app "${appName}" (disabled)`);
+			continue;
+		}
 
-		app.commandCenter = new mage.core.cmd.CommandCenter(app);
+		exports.createApp(appName);
+	}
+};
 
-		app.commandCenter.setup(function (error) {
-			if (error) {
-				return callback(error);
-			}
 
-			appMap[appName] = app;
-			appList.push(app);
+/**
+ * Returns true if the app has been created, false otherwise
+ *
+ * @param {string} name  The name of the application
+ * @returns {boolean}
+ * @throws {Error}       If the app by the given name is not configured to exist
+ */
+exports.isAppCreated = function (name) {
+	const appsConfig = exports.getAppsConfig();
+	if (!appsConfig.hasOwnProperty(name)) {
+		throw new Error(`App "${name}" has not been configured`);
+	}
 
-			callback();
-		});
-	}, cb);
+	// If the app has been set up, it exists in the appMap
+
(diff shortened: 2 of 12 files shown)


stelcheck:
Sounds good!

AlmirKadric:
👍 Sweet!

lib/app/index.js
```diff
@@ -1,53 +1,171 @@
...
 
 /**
- * This function creates instances of all configured apps and registers them into appMap and appList.
+ * Returns the configuration for all apps (overridden during unit tests)
+ *
+ * @returns {Object}
+ */
+/* istanbul ignore next */
+exports.getAppsConfig = memoize(function () {
+	return require('lib/mage').core.config.get(['apps']);
```
AlmirKadric: How is this any different to just calling the function directly?
doesn't node cache the require, and doesn't the config system cache config entry?
Is it just to avoid the object lookups?

ronkorving: A require call is still relatively expensive. Paths have to be resolved for example to find the registered (or not) module. Memoizing this stuff just makes this much cheaper to call.

ronkorving: I should add, I chose this memoization library for its [performance claims](https://github.com/medikoo/memoizee#benchmarks), but even then I'm sure we could optimize further (since we don't memoize things that receive arguments in these APIs). Right now, I've not prioritized this though. First we needed to establish this unit testing pattern. How we memoize is not as relevant, we can always swap that for something else.

lib/app/Readme.md
```diff
@@ -8,12 +8,20 @@ In your configuration, you can enable an app using something like this:
...
 apps:
-    # Replace appName with the name of the app you wish to enable
-    appName:
-        responseCache: 180
-        disabled: false
+  # Replace appName with the name of the app you wish to enable
+  appName:
+    responseCache: 180
+    disabled: false
+    sharedData:
```
stelcheck: Why do we need this for a PR that takes care of making setup synchronous?

lib/app/index.js
```diff
@@ -1,53 +1,184 @@
-var async = require('async');
+'use strict';
```
stelcheck: A module attribute and file description as a header would be most welcome.

lib/app/index.js
```diff
@@ -1,53 +1,184 @@
...
+ * @returns {Object}
+ */
+/* istanbul ignore next */
+exports.createCommandCenter = function (app) {
+	const CommandCenter = require('lib/mage').core.cmd.CommandCenter;
+	return new CommandCenter(app);
+};
+
+
+class App {
```
stelcheck: A classdesc would be most welcome.

lib/commandCenter/commandCenter.js
```diff
@@ -383,24 +373,24 @@ CommandCenter.prototype.getCommandInfo = function (cmdName) {
...
- * If two many parameters are provided, an exception will be thrown.
+ * If the params are an Array, the missing parameters will be added with undefined as value,
+ * so that the built parameter list will have the expected length.
+ * If too many parameters are provided, an exception will be thrown.
  *
- * @param {string[]}    cmdInfoModParams  Parameters information from the user module
- * @param {Object}      cmdParams         Parameters to pass to the user command function
- * @returns {Array}     Built parameter list
- * @throws  {Error}     Will throw an error if there are too many parameters provided.
+ * @throws {Error}                     Will throw an error if there are too many parameters provided.
```
stelcheck: nice

lib/app/index.js
```diff
@@ -66,11 +196,10 @@ exports.get = function (name) {
 /**
  * Returns all app instances in a list
```
stelcheck: Small detail, but I would perhaps take the opportunity to say that you now return a copy of the list, not the list itself (just to clarify the use of `slice()` here)

lib/commandCenter/commandCenter.js
```diff
@@ -1,39 +1,39 @@
...
-var mage;
-var logger;
-var State;
-
-var DEFAULT_RESPONSE_CACHE_TTL = 3 * 60;  // cache lifetime of the command response in seconds
-var COMPRESSION_THRESHOLD = 4096;         // only gzip if the response is at least this many bytes uncompressed
-var COMMANDS_DIR_NAME = 'usercommands';     // user command directory inside of module file space
-var COMMAND_PATH_SPLITTER = '.';                // what separates the module name and command name
-var CONTENT_TYPE_JSON = 'application/json; charset=UTF-8';
+'use strict';
```
stelcheck: This file has changed enough that the diff does not automatically load on Github. Are the current unit tests for commandCenter covering the methods that have been touched here?

lib/app/index.js
ronkorving: Indeed. I think that should be our default approach in all getter functions. You don't really wanna give something back that when manipulated without ill-intent would break the behavior. `slice` is also a very cheap operation in JS. I'm not sure why, but I suspect it's copy-on-write.

lib/app/Readme.md
ronkorving: We always had this, but it was undocumented. I took the opportunity to clarify it. And because it was undocumented, I took the liberty to name it better (imho).

lib/commandCenter/commandCenter.js
ronkorving: Was already there ;)

lib/app/Readme.md
stelcheck: TBH I am not sure why this name  is good, however; we are not really sharing data, but simply passing additional configuration to user-land, if I get this correctly.

lib/commandCenter/commandCenter.js
ronkorving: Unfortunately, coverage on this module is still rather poor. It has increased by this PR according to coveralls, but not by much.

stelcheck: If I may rephrase, are we at least already testing the public methods that have been modified by this PR?

ronkorving: Not yet, but I will make it so 👍

ronkorving: Alright, coverage jumped significantly now. The only uncovered bits are those where user commands are pulled from disk. But the logic behind registering the user commands is covered.

lib/app/Readme.md
ronkorving: The big question is: what defines configuration? You can really send anything you like, and it doesn't have to fall under the name "configuration". It's just app-specific "stuff".

lib/commandCenter/commandCenter.js
```diff
@@ -1,39 +1,39 @@
...
+
+let mage;
+let logger;
+let State;
 
-function UserCommandSetupError(message, details) {
-	Error.captureStackTrace(this, this.constructor);
-	Object.assign(this, details);
+const DEFAULT_RESPONSE_CACHE_TTL = 3 * 60;  // cache lifetime of the command response in seconds
+const COMPRESSION_THRESHOLD = 4096;         // only gzip if the response is at least this many bytes uncompressed
```
AlmirKadric: probably not the scope of this PR but making some of these configurable might be nice?

AlmirKadric:
LGTM 👍

lib/app/index.js
AlmirKadric: sounds good to me 👍

lib/commandCenter/commandCenter.js
ronkorving: Compression threshold maybe yes. All other constants are constants out of stylistic consideration (I don't like to duplicate them in multiple places in code). But indeed, not in the scope of this PR.

lib/app/index.js
ronkorving: Done

lib/app/Readme.md
AlmirKadric: Wow i didn't even know that existed, however i can see its merits
After a chat with @ronkorving, I agree that since we use external build systems now that can inject extra configuration, we no longer need something like this.

ronkorving: Spoken offline, we all agreed on removal.

stelcheck:
OMGWTFBBGLGTM Friday Special ✨

lib/httpServer/transports/http/index.js
```diff
@@ -57,24 +57,28 @@ exports.setCorsConfig = function (config) {
...
-			throw new TypeError('Expose URL is not a string: ' + config);
-		}
+	if (typeof url !== 'string') {
+		throw new TypeError('Expose URL is not a string: ' + url);
 	}
 
-	expose = config;
+	// drop trailing slashes (for later path concatenation)
+
+	while (url.slice(-1) === '/') {
```
qubyte: Tiny gripe. In Node 4+ you can do `url.endsWith('/')`.

test/apps/index.js
```diff
@@ -0,0 +1,192 @@
...
+			return {
+				setup: function () {
+					currentCommandCenterSetup(app);
+				}
+			};
+		}
+	};
+
+	beforeEach(function () {
+		libApp.getAppsConfig = testFns.getAppsConfig;
```
qubyte: Maybe not something for right now, but food for thought. I've found [sinon](http://sinonjs.org/) extremely useful for spying and stubbing. It avoids a lot of boiler plate and even helps me to formulate tests. `oldFns` and `testFns` above can be avoided by doing something like:

```javascript
const sinon = require('sinon');
const sandbox = sinon.sandbox.create();

before(() => {
    sandbox.stub(libApp, 'getAppsConfig').returns(currentAddsConfig);
    sandbox.stub(libApp, 'getLogger').returns(() => {
        return { debug: sandbox.stub() };
    }));
    sandbox.stub(libApp, 'createCommandCenter').returns(app => {
        return {
            setup() {
                currentCommandCenterSetup(app);
            }
        };
    }));
});

afterEach(() => {
    sandbox.reset(); // Reset stub call counts, args, etc.
    libApp.removeAllApps();
});

after(() => {
    sandbox.restore(); // Unstub.
});
```

The above is more like a mock, so you don't really care about inputs or call order. Stubs do record those though, which can be really handy. The sandbox here isn't strictly necessary, but it helps to ensure that stubbed methods aren't forgotten and leak into other tests.

ronkorving: I'll have a look, thanks 👍 
My general attitude is one where I'm trying as much as possible to keep things simple (that is: you read the code and you know immediately what it does, because there's ideally no library in between). But at some point, a library becomes useful enough to convince me otherwise ;)

PR #5, opened 2017-04-22
typescript support & type description

stelcheck:
### Todo

  - [x] Complete the description file
  - [x] Documentation update
  - [x] Update Typescript template to use `import` instead of `require`
  - [x] Write some test code!

The definition file will be picked up both in JavaScript and Typescript
projects.

The project templates have been split into two so that we may actively
support Typescript-based MAGE projects.


christopherobin:
You should use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) in your `.d.ts` file to make types available, something like:

```typescript
declare var mage: mage.IMage;

declare namespace mage {
  interface IMage {
    // the actual mage object def
  }

  interface IState {
  }
}

export = mage;

// note: a namespace can only be merged with a var if it
// contains only types and interfaces; using class or abstract
// class will prevent the compiler from merging
```

Another good example: [DefinitelyTyped/DefinitelyTyped/types/angular/index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/angular/index.d.ts)

This will allow users to have access to both the methods on `mage` and also access the types themselves, eg.

```typescript
import mage = require('mage');

export function execute(state: mage.IState, cb: mage.UserCommandCallback) {
  state.respond(mage.listModules());
  cb();
}
````

See this quick&(very)dirty declaration file I wrote over the weekend: https://gist.github.com/christopherobin/9969d18277de74b218f9ca7007906d0a

**Comment by stelcheck** [comment]
@christopherobin I hadn't seen your comment, but could you look at the latest version of the file?

I did it the way it is now because with declaration merging I had issues with either namespace methods not being available or object instance methods not being available. Not sure if that was a bug or an issue, but I think merging works with class and namespaces, not object *instances* and namespaces.

File: docs-sources/includes/_Installation.md [hunk 1]
```diff
@@ -44,12 +46,27 @@ npm run develop
...
 
-Running the following steps is the easiest way to bootstrap a new project. Do this **from inside an
+Running the following steps is the easiest way to create a new project. Do this **from inside an
 empty folder** that is named after the game you are developing.
 
-You can also specify which version you wish to install by adding `#[version number]` at
+You can also specify which version you wish to install by adding `@[version number]` at
 the end of the line.
 
+### As a Typescript project
```
ronkorving: TypeScript

types.d.ts
```diff
@@ -0,0 +1,1003 @@
...
+    }
+}
+
+declare var mage: Mage;
+
+declare namespace mage {
+    namespace archivist {
+        interface IVault {
+
+	    }
```
ronkorving: indentation is off

types.d.ts
```diff
@@ -0,0 +1,1003 @@
...
+
+        interface IModule {
+            setup?(state: IState, cb: (error?: Error) => void): void;
+            start?(state: IState, cb: (error?: Error) => void): void;
+            shutdown?(state: IState, cb: (error?: Error) => void): void;
+        }
+
+        export interface IState {
+            /**
+             * Archivist instance
```
ronkorving: ?

stelcheck: What is your question?

stelcheck:
Pushed an update to the definition file. 

I left a whole bunch of Todos in it, wherever things need to get completed. 

I would need a bit of help with the following:

  - Completing todos
  - Testing (since its perfectly possible to write code that compiles but crashes at runtime because of discrepancies between the existing code and the type definition files)

Help most welcome @ronkorving @AlmirKadric @christopherobin

types.d.ts
ronkorving: Nevermind, I thought you made an "obvious" copy/paste error. Instead, I made a reading error. Problem is between keyboard and user.

types.d.ts
```diff
@@ -0,0 +1,1453 @@
...
+     *
+     * @param {string} topicName
+     * @param {ArchivistIndex} index
+     * @param {ArchivistOptions} options
+     * @param {Callback} callback
+     *
+     * @memberOf Archivist
+     */
+    get(topicName: string, index: ArchivistIndex, options: ArchivistGetOptions, callback: ArchivistGetCallback): void;
+    get(topicName: string, index: ArchivistIndex, callback: ArchivistGetCallback): void;
```
christopherobin: It would be nice to have:

```typescript
type TypedArchivistGetCallback<T> = (err: Error, val: T) => void;
get<T>(topicName: string, index: ArchivistIndex, callback: TypedArchivistGetCallback<T>): void;
```

So that in my code I can do:

```typescript
state.archivist.get<IGame>("game", { gameId: "gameId" }, (err, val) => {
  // val is an IGame instance there
});
```

stelcheck: Indeed, will add it

stelcheck:
Type definition has been completed (thanks to @ronkorving for the help)]

The code is now almost ready for merge. Last step is to:

  1. Make sure we get compile errors wherever it make sense
  2. Make sure that valid code compiles properly
  3. Make sure that compiled code does not crash at runtime due to type definition file discrepancies or other issues

@AlmirKadric @kefniark @ronkorving @christopherobin @nullorvoid Comments and test results welcome!

lib/commandCenter/commandCenter.js
```diff
@@ -198,7 +199,7 @@ CommandCenter.prototype.setupUserCommand = function (modName, cmdName, cmd) {
 	const firstParameter = params.shift();
 
 	// Sanity check
-	if (firstParameter !== 'state') {
+	if (params.length !== 0 && firstParameter !== 'state') {
```
stelcheck: Fixes a bug where in async user commands, you may not even need to use the state object

lib/mage/Mage.js
```diff
@@ -60,13 +71,18 @@ function Mage(mageRootModule) {
 
 	try {
 		rootPackage = require(rootPackagePath);
-	} catch (e) {}
+	} catch (error) {
```
stelcheck: Fixes an issue where if your project's package.json is corrupted, it won't load and you will not be notified.

scripts/install.js
```diff
@@ -1,11 +1,17 @@
 #!/usr/bin/env node
 
-var npmArgs = JSON.parse(process.env.npm_config_argv);
+var npmArgs = JSON.parse(process.env.npm_config_argv).original;
 
-if (npmArgs.original.indexOf('--bootstrap') === -1) {
+if (npmArgs.indexOf('--create') === -1 && npmArgs.indexOf('--bootstrap') === -1) {
```
stelcheck: `--create` is shorter to type and more explicit I find. For now, we can support both, but would suggest we start moving towards `--create` (adjusted the documentation accordingly)

docs-sources/includes/_Installation.md
```diff
@@ -44,12 +46,27 @@ npm run develop
 npm run develop
 ```
 
-Running the following steps is the easiest way to bootstrap a new project. Do this **from inside an
+Running the following steps is the easiest way to create a new project. Do this **from inside an
 empty folder** that is named after the game you are developing.
 
-You can also specify which version you wish to install by adding `#[version number]` at
+You can also specify which version you wish to install by adding `@[version number]` at
```
ronkorving: Nice catch

lib/mage/Mage.js
```diff
@@ -91,6 +107,29 @@ function Mage(mageRootModule) {
...
+ *
+ * It is also noteworthy that in the case of TypeScript, if you
+ * choose to manually transpile your project to JavaScript, you
+ * should then set the project type to JavaScript instead; if you
+ * are using 'ts-node' (this is default for bootstrapped MAGE TypeScript
+ * projects), then the type should be TypeScript.
+ *
+ * @summary Set the type of project (JavaScript or TypeScript)
+ * @param {string} projectType Type of project ('javascript' or 'typescript')
+ * @returns {Mage} This MAGE instance
```
ronkorving: You can add a `@throws` entry

scripts/install.js
ronkorving: Absolutely agree 👍

scripts/install.js
```diff
@@ -236,6 +262,11 @@ function bootstrap(cb) {
 			});
 		},
 		function (callback) {
+			// For some reason, when using the --prefix flag, an etc folder is added.
```
ronkorving: Any idea where this is coming from? Is that NPM, and if so, is there an issue for this?

lib/mage/Mage.js
```diff
@@ -91,6 +107,30 @@ function Mage(mageRootModule) {
...
+ * @throws {Error} Thrown if the project type is unknown
+ */
+Mage.prototype.setProjectType = function (projectType) {
```
ronkorving: I'm not too thrilled about the name of this method. Not sure I can give any better alternatives though, but a "project type" could be anything (eg: "game").

scripts/install.js
stelcheck: > global configuration file (defaults to $PREFIX/etc/npmrc; configurable via CLI option --globalconfig or environment variable $NPM_CONF_GLOBALCONFIG)

From what I can see from the NPM history, it creates it preemptively: https://github.com/npm/npm/pull/7249

scripts/templates/typescript-project/config/default.yaml
```diff
@@ -0,0 +1,109 @@
...
+# individual environments).
+
+apps:
+    # Our main app: the game
+    game:
+        # Number of seconds to keep user command responses cached (to mitigate weak client connections)
+        responseCache: 30
+
+        # The maximum access level for this app (user commands with a higher level are not exposed)
+        access: user
```
stelcheck: I did. This is the typescript template, not the javascript one, and this was started prior to your other PR being made. Will fix it.

lib/mage/Mage.js
stelcheck: That's what documentation is for :)

bin/ts-mage
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env ts-node
+
+// We only need this file to specify that
+// we should run using ts-node instead of node.
+
+// Require the node version
```
ronkorving: The node version?

stelcheck: The node version of this binary (which sits right next to it)

lib/mage/Mage.js
```diff
@@ -91,6 +98,20 @@ function Mage(mageRootModule) {
...
+/**
+ * Example:
+ *
+ * ```javascript
+ * mage.isCodeFileExtension('.js');
+ * ```
+ *
+ * @summary Check if a file is considered like a source code file in MAGE
+ * @param {string} ext File extension
+ * @returns {boolean} This MAGE instance
```
ronkorving: Not "this MAGE instance" :)

bin/ts-mage
ronkorving: I get what you mean now. Maybe change "node" to "JavaScript" or "non-TypeScript"?

scripts/install.js
stelcheck: True, they might introduce bootstrap as well - at any rate, for now, we have a fallback :D

scripts/install.js
ronkorving: Maybe add that issue URL to the code comment?

scripts/install.js
```diff
@@ -1,11 +1,17 @@
...
 
-if (npmArgs.original.indexOf('--bootstrap') === -1) {
+if (npmArgs.indexOf('--create') === -1 && npmArgs.indexOf('--bootstrap') === -1) {
 	process.exit(0);
 }
 
+var language = 'javascript';
+
+if (npmArgs.indexOf('--typescript') !== -1 || npmArgs.indexOf('--ts')  !== -1) {
+	language  = 'typescript';
```
ronkorving: Nitpicky, but you have an extra space after `language`.

scripts/install.js
```diff
@@ -1,11 +1,17 @@
...
+var npmArgs = JSON.parse(process.env.npm_config_argv).original;
 
-if (npmArgs.original.indexOf('--bootstrap') === -1) {
+if (npmArgs.indexOf('--create') === -1 && npmArgs.indexOf('--bootstrap') === -1) {
 	process.exit(0);
 }
 
+var language = 'javascript';
+
+if (npmArgs.indexOf('--typescript') !== -1 || npmArgs.indexOf('--ts')  !== -1) {
```
ronkorving: Nitpicky, but you have an extra space after the 2nd `indexOf()` call.

scripts/install.js
stelcheck: Added

ronkorving:
LGTM

PR #7, opened 2017-04-24
Travis CI and Coveralls

ronkorving:
This adds a config file for Travis and integration with coveralls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,9 @@
 language: node_js
 node_js:
   - 4
+  - 5
   - 6
   - 7
-
-script: npm test
-
-env:
-  global:
-    - NODE_ENV=development
-    - CXX=g++-4.8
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -30,6 +17,27 @@ before_install:
   - make install
   - cd ..
 
+script:
```
stelcheck: Can't we paralellize that? https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

ronkorving: Yes, but they won't share ZMQ installations. That takes up a lot of time.

ronkorving: Also, it runs coveralls for all versions of Node then, which is a little silly.

But, it *is* a lot more elegant to split it up, so I reverted this.

stelcheck:
Rebased an merged manually into master

PR #14, opened 2017-04-24
Enhances coverage of config and removes a few redundant code paths.

qubyte:
It looked like my patchy old unit tests could use some enhancements, so that's exactly what I did. This is still rather flat and haphazard though, so if you'd prefer testing of methods in different sub-describes then let me know.

### DIFF
--- a/lib/config/Matryoshka/index.js
+++ b/lib/config/Matryoshka/index.js
@@ -140,12 +140,6 @@ Matryoshka.prototype.tunnel = function (path) {
  */
 
 function getRaw(matryoshka) {
-	var isMatryoshka = matryoshka instanceof Matryoshka;
-
-	if (!isMatryoshka) {
-		return matryoshka;
-	}
-
 	if (matryoshka.type === 'object') {
 		var returnObj = {};
 		var keys = Object.keys(matryoshka.value);
@@ -273,10 +267,6 @@ Matryoshka.prototype.copy = function () {
  */
 
 function mergeObjects(a, b) {
-	if (a.type !== 'object' || b.type !== 'object') {
-		throw new TypeError('Arguments must be (non-null, non-array) object containing Matryoshka.');
-	}
-
 	var joinKeys = Object.keys(a.value);
 	var bKeys = Object.keys(b.value);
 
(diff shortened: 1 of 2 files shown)


qubyte:
Not sure what the AppVeyor breakage is about.

qubyte:
0.2%

Really getting stuck into the low hanging fruit. 🤣

qubyte:
By the way... Given that mage is tested on Node 4 and up, what is your policy on supported ES2015 features?

stelcheck:
The Appveyor failure is because you do not have the Appveyor config file  which is in PR #11 right now.

ronkorving:
@qubyte Hey, first of all, thanks for contributing, that's really cool :)

With regards to what language features you can use. You may use most language features the lowest common denominator (Node 4) supports (in fact, it's encouraged), which is quite a lot. I do ask in those cases that you mark the files to `'use strict';`.

What I do want to discourage is using language features "because we can". The reader of the code comes first, performance comes second, clever trickery comes.. well, it's not on the list ;)

ronkorving:
#11 has been merged by the way

qubyte:
I'll rebase on master to fix the appveyor breakage.

Agreed on all counts about usage of features! The reason I ask in this case is because there are a few places which would definitely benefit. For example, this:

```javascript
var joinKeys = Object.keys(a.value);
var bKeys = Object.keys(b.value);

// Add keys from b.value that were not present in a.value;
for (var i = 0; i < bKeys.length; i++) {
	var bKey = bKeys[i];

	if (joinKeys.indexOf(bKey) === -1) {
		joinKeys.push(bKey);
	}
}
```

is more succinctly expressed in a Node 4 and above compatible way as:

```javascript
const aKeys = Object.keys(a.value);
const bKeys = Object.keys(b.value);

// A union of unique keys from a.value and b.value.
const joinKeys = Array.from(new Set(aKeys.concat(bKeys)));
```

I can open a follow-up PR with suggestions soon.

ronkorving:
@qubyte Ah yes, that code is example is rather on the edge for now :) I don't particularly mind that remaining old-style.

qubyte:
I'd go the route of a function to name what's going on there.

qubyte:
Or, given that we don't actually need that to be an array...

```javascript
function mergeObjects(a, b) {
	const aKeys = Object.keys(a.value);
	const bKeys = Object.keys(b.value);
	const joinKeys = new Set(aKeys.concat(bKeys));
	const returnObj = {};

	// Compare each key-by-key.
	for (const key of joinKeys) {
		returnObj[key] = merge(a.value[key], b.value[key]);
	}

	// Make a new matryoshka with the results.
	return new Matryoshka(returnObj, a.source, true);
}
```

qubyte:
Ouch, sorry about that.

ronkorving:
Don't forget 'use strict' at the top of the file (or your const iteration will break on Node 4)

ronkorving:
Superseded by #17

PR #15, opened 2017-04-25
Add support for macOS in travis

### LINKED ISSUE #12: Run Travis tests on MacOS
https://docs.travis-ci.com/user/osx-ci-environment/

ronkorving:
Closes #12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+os:
+  - linux
+  - osx
+
 language: node_js
 node_js:
   - 4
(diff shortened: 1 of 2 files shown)


stelcheck:
Could you also rename the text on the badge from Linux to something like "Linux / macOS", like so:

[Build Status: Linux & macOS](https://travis-ci.org/mage/mage)

stelcheck:
Also (unrelated), a message from coveralls on each commit pushed is a bit loud. Any way we could set this to something more unreasonable, such as sending a message only when coverage goes down?

ronkorving:
It's not just each commit I think. We run coveralls for each Node version! That's one reason why at first, I had travis.yml "scripts" section run all node versions one-by-one.

stelcheck:
We should maybe make a separate job for the coverage then - and even there, I would try to make it so not to get notifications unless we have a drop in coverage. We already have GitHub integrated notification markers when tests are failing, adding an explicit message seems redundant in most cases.

ronkorving:
Drop in coverage beyond 0.5% now constitutes failure. I've turned off messages.

PR #17, opened 2017-04-25
Config node 4 enhancements

qubyte:
Some speculative changes as discussed in #14. This PR supersedes #14 if it is deemed acceptable.

Changes to Matryoshka were done against the old tests. The sole breakage is that Matryoshka can no longer created without using `new`. Since this was unused behaviour I went ahead and used a `class`.

Following those changes, I trashed most of the tests and rewrote them in a more structured way. Hopefully it makes the intended behaviour clearer.


ronkorving:
Very clean, thanks! LGTM

PR #37, opened 2017-05-18
Type definition file driven API doc

stelcheck:
### Todo

 - [x] Port content from the old API documentation to the type definition file

### Description

Requires #5 

This PR introduces Typedoc as our system
for API doc generation. Some refactoring
of the current documentation will be required (example: vault backends
are now in the API doc, but should probably be in the user doc).


stelcheck:
Is anyone seeing value in this? To be honest, after a few months of working with VSCode on MAGE projects, I have to admit I get all the docs I need straight out of my IDE.

Would love to get feedback from developers using other IDEs, so to know whether I should even pursue with this PR.

stelcheck:
Rebased the branch. Once again, would love some opinion on this; do you like how it looks? Do you see any potential issues with this approach?

stelcheck:
Completed the remainder of the work. I consider this ready to review and merge.

stelcheck:
Here you can access the content of this PR: https://stelcheck.github.io/mage/

stelcheck:
Strangely enough, there seems to be a peer dependency issue, only on Node 4 and on Travis 

https://travis-ci.org/mage/mage/jobs/256306990#L2712

stelcheck:
All tests should now be passing. See https://github.com/mage/mage/pull/37/commits/2aa0dc3e491ef3f99bb78c1f5556bda8bde0d407 for more details.

stelcheck:
@AlmirKadric @ultimatezen @christopherobin we really need some feedback on this one before merging, and I would like to merge in a timely fashion. Could you please provide comments/feedback/approval/insults/etc?

ultimatezen:
Small nitpick, but even scrolled all the way to the bottom, the menu highlighting won't hit `Production` (document height is too short?).
image

stelcheck:
@ultimatezen yeah, I think that must be it. Have you also looked at https://stelcheck.github.io/mage/api/index.html? This section is the one which gets entirely revamped by this PR (apologies for not explicitly linking to it).

ultimatezen:
Took a quick look and seems ok to me. Only gripe would be some stuff is harder to find if you don't know where to look. i.e. `State` is in `core`, `time` is on `mage`.

stelcheck:
@ultimatezen Did you notice the search box at the top-right of each page? It is not as visible as I would like it to be, guess I could try to look for/make  a custom template.

ultimatezen:
@stelcheck Noticed it after a bit of searching, but yeah, custom template could be interesting for a future improvement.

AlmirKadric:
Here is my feedback on the 2 doc systems.
I am putting all my thoughts here.
With this being said, whether these get fixed now or later, I will leave it up to you.

The API doc, as a reference looks fine, however:
 - why is mage under the globals as the root? i would try and customize that structure, it look and feel unnatural to use
   have all the subsystems sit there instead
   what they care about is the content of the mage object thus it should be the root, not global
 - Some of the more important classes (like State) are too hard to find
   Does the api doc system support pseudo class pages?
   this will help link some of the content in a more comprehensible way
 - does it support readme sections on larger classes?
   (e.g. the archivist class explanation is way too thin)
   (having to look in 2 places for this information doesn't make sense)
 - does is support localisation?
 - can we not integrate it back into the same format as the user doc?
   (moving between 2 different sites is a bit meh)

The user doc:
 - template wise is nice
 - current content and structure is not really useful for a new user of mage (IMHO)
 - the structure on the left is a bit funky (maybe think about creating a bit more nesting?)
 - the sections of text don't have a nice pretext, so it looks a bit like here is information
   (no context to why it is useful)
   however we can improve this as we improve the content
 - the fact that i have a link pointing to api doc inside the text is weird
   if we cant integrate it the link on the side should open a new page

stelcheck:
> why is mage under the globals as the root? i would try and customize that structure, it look and feel unnatural to use

Already tried, but this is the best I was able to come up with. I agree this is not super great, but we could try to fix this in the future I think.

As for the other comments, I would suggest we address them in later improvements to the doc. For one, I would indeed like to have a more consistent UI style between the user doc and the API doc.

stelcheck:
Was able to address some of the issues; asking around one last time (specifically about the switch to a minimal theme - looks better, but no search box).

AlmirKadric:
looks much better 👍 
lets fix the rest later, we can move the above comments into another issue

PR #38, opened 2017-05-18
Support proper versioning through NPM

ronkorving:
This allows "npm version <version/major/minor/patch>" to do its thing, while generating uptodate docs for the release commit. It also includes a release guide to document the entire release process.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,3 +2,5 @@ heading = mage
 progress = false
 loglevel = http
 save-prefix = ""
+message = "Release %s"
+tag-version-prefix = ""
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -2,3 +2,5 @@ heading = mage
 progress = false
 loglevel = http
 save-prefix = ""
+message = "Release %s"
```
ronkorving: Means the commit message will be "Release 1.2.3"

.npmrc
```diff
@@ -2,3 +2,5 @@ heading = mage
...
+tag-version-prefix = ""
```
ronkorving: Means when NPM tags the git commit, it will not prefix a "v".

PR #41, opened 2017-05-20
state.emit/broadcast eventName can be number, type

stelcheck:
This is to allow better type analysis by the parser;
it also allows for developers to explicitly type the call,
as follow:

```typescript
state.broadcast<string>('hello', 'world')
```

Also, allowing for the eventName to be a number allows
for the following code pattern:

```typescript
// lib/modules/test/index.ts
import * as mage from 'mage'

export const enum events {
  Success = 5000,
  Failure,
  Dunno
}

export function emitSuccessValue(actorId: string, state: mage.core.IState) {
  state.emit(actorId, events.Success, 1)
}
```

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1431,22 +1431,22 @@ declare namespace mage {
              * is not currently connected.
              *
              * @param {(string | string[])} actorId
-             * @param {string} eventName
+             * @param {string | number} eventName
              * @param {*} data
              *
              * @memberOf State
              */
-            emit(actorId: string | string[], eventName: string, data: any): void;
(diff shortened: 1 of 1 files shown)


stelcheck:
Note: type variables are optional as long as the type of the variable can be inferred; the following would still be valid

```typescript
state.broadcast('hello', 'world')
```

ronkorving:
The Web SDK explicitly expects event names to be strings. The event manager will emit 3 events for an event named `"foo.bar.helloworld"`: 

1. `"foo.bar.helloworld"`
2. `"foo.bar"`
3. `"foo"`

So allowing numbers would change things quite a bit.

The change to use a template type for `data` is LGTM 👍

PR #42, opened 2017-05-21
IUserCommand.execute type

stelcheck:
Enforce type on usercommand.execute. This will:

  1. Pick up if the `async` keyword is missing
  2. Allow for explicity typedefinition if desirable

Note that while MAGE supports callback-based user commands,
they would now become invalid for TypeScript; the reason for this
that the only valid type definition we could introduce is the following:

execute<T>(state: IState, ...args: any[]): T;

Which would:

  1. Not check that the last parameter is an empty callback function
  2. Essentially make checking for the `async` keyword impossible
  (because of the overloading)

Since we are promoting the use of `async` anyway, I figured out this
wouldn't matter; beside, if a developer really wanted to use
callback-based execute function, they could always write it like
without typing.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1362,7 +1362,7 @@ declare namespace mage {
          */
         interface IUserCommand {
             acl?: string[];
-            execute(state: IState, ...args: any[]): any;
+            execute<T>(state: IState, ...args: any[]): Promise<T>;
         }
 
         export interface IState {


stelcheck:
The only real issue I have experience so far with this change is that now, somehow, the type of `state` is no longer inferred:

```typescript
export = <mage.core.IUserCommand> {
    acl: ['*'],
    async execute(state: mage.core.IState) {
        test.emitSuccess(state)
        return new B('asd')
    }
}
```

I am not entirely sure why that is, but I personally don't mind too much having to add the extra type information in this case.

types.d.ts
```diff
@@ -1362,7 +1362,7 @@ declare namespace mage {
          */
         interface IUserCommand {
             acl?: string[];
-            execute(state: IState, ...args: any[]): any;
+            execute<T>(state: IState, ...args: any[]): Promise<T>;
```
micky2be: Try `execute(state: IState, ...args: any[]): Promise<any>;` instead.

stelcheck: I actually want to use the type argument to explicitly type (or infer) the return value of the execute method.

PR #44, opened 2017-05-21
module index file not required

### LINKED ISSUE #35: Modules require an 'index.js'
I may have a module composed only of usercommands; in which case, creating an empty index.js at the top level of the module seems to be a bit of a waste.

stelcheck:
In projects, the following structure is now OK:

```plaintext
lib/
  modules/
    moduleWithOnlyUserCommands/
      usercommands/
        a.js
        b.js
        c.js
```

Notice that there are no `index.js` in the module (in which
case, a warning log will be printed out nevertheless)

Fixes #35

### DIFF
--- a/lib/mage/Mage.js
+++ b/lib/mage/Mage.js
@@ -410,13 +410,23 @@ Mage.prototype.useModules = function () {
 		this._modulesList.push({ name: name, path: resolved });
 		this._setupQueue.push(name);
 
+		var mod = {};
+
 		try {
-			var mod = require(resolved);
-			this[name] = this.core.modules[name] = mod;
+			mod = require(resolved);
 		} catch (error) {
-			error.module = name;
-			error.moduleFile = resolved;
-			throw error;
+			if (error.code !== 'MODULE_NOT_FOUND') {
+				error.module = name;
+				error.moduleFile = resolved;
+				throw error;
+			}
+
+			this.core.logger.warning.data({
+				moduleName: name,
+				modulePath: resolved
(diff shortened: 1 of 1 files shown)


lib/mage/Mage.js
```diff
@@ -410,13 +410,23 @@ Mage.prototype.useModules = function () {
...
+			if (error.code !== 'MODULE_NOT_FOUND') {
+				error.module = name;
+				error.moduleFile = resolved;
+				throw error;
+			}
+
+			this.core.logger.warning.data({
+				moduleName: name,
+				modulePath: resolved
+			}).log('No module found for module');
```
ronkorving: That's a bit of a funky way to express this error :)

"No module found for module"

stelcheck: Hey, hey... its Sunday.

lib/mage/Mage.js
```diff
@@ -410,13 +410,23 @@ Mage.prototype.useModules = function () {
...
+			if (error.code !== 'MODULE_NOT_FOUND') {
+				error.module = name;
+				error.moduleFile = resolved;
+				throw error;
+			}
+
+			this.core.logger.warning.data({
+				moduleName: name,
+				modulePath: resolved
+			}).log('No module source entrypoint found');
```
stelcheck: I can't find a better log entry text at this time. Suggestions welcome @ronkorving

ronkorving: Web SDK implementation:

```js
console.warn('Module "' + name + '" has no implementation.');
```

stelcheck:
Done

PR #45, opened 2017-05-21
savvy socket file config: absolute paths

### LINKED ISSUE #26: Cannot create savvy sock file outside of process.cwd()
https://github.com/mage/mage/tree/45b7de63797389bc4ed011ee105f7ec2396bcaae/lib/savvy/index.js#L18

This is an issue in my case (running in docker-compose):

1. The file doesn't get cleaned up; so I was hoping to put that socket file directly in the container's filesystem instead
2. More a general case, but it looks a bit strange to enforce the extension on this one

stelcheck:
With this fix, it should now be possible to specify absolute
paths in your configuration.

Fixes #26

### DIFF
--- a/lib/savvy/index.js
+++ b/lib/savvy/index.js
@@ -10,14 +10,12 @@ mage.core.config.setTopLevelDefault('savvy', path.join(__dirname, 'config.yaml')
 var ROUTE_STR = '/savvy';
 var ROUTE_RE = new RegExp('^' + ROUTE_STR + '/');
 
-var SOCK_PATH;
 var SOCK_FILE = mage.core.config.get(['savvy', 'sockfile']);
-if (process.platform === 'win32') {
-	SOCK_PATH = path.join('\\\\.\\pipe', process.cwd(), SOCK_FILE);
-} else {
-	SOCK_PATH = path.join(process.cwd(), SOCK_FILE + '.sock');
-}
+var SOCK_PATH = path.resolve(SOCK_FILE);
 
+if (process.platform === 'win32') {
+	SOCK_PATH = path.join('\\\\.\\pipe', SOCK_PATH);
+}
 
 function trimRegExp(re) {
 	re = re.toString();
--- a/lib/savvy/config.yaml
+++ b/lib/savvy/config.yaml
@@ -1,1 +1,1 @@
-sockfile: "savvy"
+sockfile: "savvy.sock"


lib/savvy/index.js
```diff
@@ -10,14 +10,17 @@ mage.core.config.setTopLevelDefault('savvy', path.join(__dirname, 'config.yaml')
...
 
-var SOCK_PATH;
 var SOCK_FILE = mage.core.config.get(['savvy', 'sockfile']);
-if (process.platform === 'win32') {
-	SOCK_PATH = path.join('\\\\.\\pipe', process.cwd(), SOCK_FILE);
-} else {
-	SOCK_PATH = path.join(process.cwd(), SOCK_FILE + '.sock');
+
+if (!path.isAbsolute(SOCK_FILE)) {
+	SOCK_FILE = path.resolve(process.cwd(), SOCK_FILE);
```
AlmirKadric: I think `process.cwd()` is redundant here. `path.resolve()` will always return an absolute path https://nodejs.org/api/path.html#path_path_resolve_paths. Also due to its behavior, you could avoid using the `path.isAbsolute()` check.

lib/savvy/index.js
```diff
@@ -10,14 +10,17 @@ mage.core.config.setTopLevelDefault('savvy', path.join(__dirname, 'config.yaml')
...
 }
 
+var SOCK_PATH = SOCK_FILE + '.sock';
```
AlmirKadric: you don't need the .sock on windows, this is why originally it was only in the non-win32 branch of the condition.

lib/savvy/index.js
```diff
@@ -10,14 +10,17 @@ mage.core.config.setTopLevelDefault('savvy', path.join(__dirname, 'config.yaml')
...
+
+if (process.platform === 'win32') {
+	SOCK_PATH = path.join('\\\\.\\pipe', SOCK_FILE);
```
AlmirKadric: we use the `process.cwd()` in the original code to avoid conflicts and keep names pacing similar between platforms.

lib/savvy/index.js
AlmirKadric: Also considering your comment in the original issue, why not just scrap this line all together?

AlmirKadric:
Personally I would have refactored this as follows to resolve your issue and simplify the logic but keeping the name spacing:
```
const SOCK_FILE = mage.core.config.get(['savvy', 'sockfile']);
let SOCK_PATH = path.resolve(SOCK_FILE);
if (process.platform === 'win32') {
    SOCK_PATH = path.join('\\\\.\\pipe', SOCK_PATH);
}
```

This would also remove junk `../../../` etc. from the name spacing on win32 platforms.

lib/savvy/index.js
stelcheck: The .sock won't be present on windows.

The extension added at the end is nice on linux systems I think, but I don't mind either way.

stelcheck:
Make sense to me, looks cleaner (but even in the current format, path on windows would not have been relative, however)

lib/savvy/index.js
AlmirKadric: Ah right, you use `SOCK_FILE` for windows and not `SOCK_PATH`. I too don't mind either way either. Was merely referencing what you mentioned in your issue
https://github.com/mage/mage/issues/26#issue-227301655: `More a general case, but it looks a bit strange to enforce the extension on this one`

AlmirKadric:
It would since you attach `SOCK_FILE` to `\\\\.\\pipe`. So if your configured socket path is relatively higher (what was the issue in the first place) you would get the following `\\\\.\\pipe\..\..\some_socket`

AlmirKadric:
LGTM 👍

PR #54, opened 2017-05-24
Moving Archivist types into mage.archivist

stelcheck:
Additionally, also making the expirationTime on
certain calls optional.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,30 +1,5 @@
 /// <reference types="node" />
 
-/**
- * The media type defines how the data will be deserialized when read from a topic, and
- * how it will be serialized before it is written to a topic.
- */
-declare type ArchivistMediaType = 'application/json' | 'application/octet-stream' | 'application/x-tome' | 'text/plain' | string;
-
-/**
- * Defines to what type of data structure a topic instance will be deserialized into. It is also used
- * in some cases to specify to a serialization method what is the data type of the data we are feeding in.
- */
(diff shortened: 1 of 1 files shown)


stelcheck:
@ronkorving question: are mediaType and encoding optional parameters?

types.d.ts
```diff
@@ -1042,6 +1017,31 @@ declare var mage: Mage;
...
+        type ArchivistEncoding = 'utf8' | 'buffer' | 'live';
+
+        /**
+         * Callback functions
+         */
+        type ArchivistExistsCallback = (error: Error|null, doesValueExist: boolean) => void;
+
+        type ArchivistGetCallback<T> = (error: Error|null, value: T) => void;
+
+        type ArchivistMGetCallback<T> = (error: Error|null, value: T[]) => void;
```
ronkorving: This can also be a key/value object, not just an array. Depends on the input to mget basically.

stelcheck: Can you give me code samples?

ronkorving:
```js
const queries = {
  user: { topic: 'user', index: [userId] },
  friend: { topic: 'user', index: [friendId] }
};

archivist.mget(queries, (error, values) => {
  // values.user
  // values.friend
});
```

stelcheck: That then means we need to also create two types of queries - we only have one in the type def.

Are there official names for those two different types of queries?

ronkorving: I don't believe there are.

PR #59, opened 2017-05-31
Allow Node 8

ronkorving:
This also removes Node 5 testing from Travis and AppVeyor, as it's been end-of-life for a little while now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,14 +10,14 @@
 ._.DS_Store
 .htpasswd
 .pidfile
-npm-debug.log
 .digest-cache.json
 check.txt
 .vagrant
 *.heapsnapshot
 
-# YARN local install
+# package managers
 .yarn
+npm-debug.log
 
 # Retire cache
 .retire-cache
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "types": "./types.d.ts",
   "engines": {
-    "node": "^4 || ^5 || ^6 || ^7",
+    "node": "^4 || ^6 || ^7 || ^8",
     "typescript": "^2.2"
   },
   "scripts": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ dist: trusty
 language: node_js
 node_js:
   - 4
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -10,14 +10,15 @@
...
 .digest-cache.json
 check.txt
 .vagrant
 *.heapsnapshot
 
-# YARN local install
+# package managers
 .yarn
+npm-debug.log
+package-lock.json
```
stelcheck: Are you not supposed to leave lock files in the repo?

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "types": "./types.d.ts",
   "engines": {
-    "node": "^4 || ^5 || ^6 || ^7",
+    "node": "^4 || ^5 || ^6 || ^7 || ^8",
```
stelcheck: If we do not test for 5 we should remove it from supported versions. However, since we still have known projects using 5, I would be cautious about this.

.gitignore
ronkorving: I would expect so in the application project, but in MAGE?

package.json
ronkorving: Those known projects do not use the open source version of MAGE, so I would really rather stick to non end-of-life versions of Node (and not have to support those anymore). If those projects wanna use this version of MAGE, they can in the same breath bump the version of Node they're on, right?

.gitignore
stelcheck: If we consider using them eventually... in most locking systems I have seen (mostly elixir and go vendor files), you would def. want to put them in your repo, hence my confusion.

package.json
stelcheck: Fine by me. Nuke Node 5.

.gitignore
ronkorving: For the record: https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md

What I take from this is:

- The top level project will not use it.
- `npm publish` will ignore it.
- But for MAGE development, it can be considered useful.
- So yes, commit it.

PR #61, opened 2017-05-31
ts: default export/require default

stelcheck:
The goal of this commit is to allow for user commands
and module definition through the `export default` TypeScript
syntax, which is a bit cleaner than `export =`.

@ronkorving I just dropped the code like that because honestly, I am not entirely sure how to approach this one - but overall this would do what I want.

### DIFF
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -156,6 +156,12 @@ CommandCenter.prototype.setupUserCommandFromDisk = function (modName, modPath, c
  * @param {Object} cmd       The full user command implementation (ie: the result of calling require(userCommandFile))
  */
 CommandCenter.prototype.setupUserCommand = function (modName, cmdName, cmd) {
+	// This is required for TypeScript project where you might want to
+	// be able to `export default [user command code]`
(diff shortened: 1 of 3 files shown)


lib/commandCenter/commandCenter.js
```diff
@@ -171,7 +177,7 @@ CommandCenter.prototype.setupUserCommand = function (modName, cmdName, cmd) {
 		throw new UserCommandSetupError('Command name must be a string', logDetails);
 	}
 
-	if (!cmd || typeof cmd !== 'object') {
+	if (!cmd || (typeof cmd !== 'object' && typeof cmd !== 'function')) {
```
stelcheck: Needed for cases where you wish to make it into a class with static attributes.

ronkorving: Could you add a comment stating why the `function` type is valid? Mentioning classes that expose `execute` and `acl` statically.

ronkorving:
Could you add a unit test for this at https://github.com/mage/mage/blob/master/test/commandCenter.js#L38 ?

Testing both the passing of a function instead of an object, and exporting a `default` property.

PR #62, opened 2017-06-07
archivist.loaded

stelcheck:
Exposing this will allow developers to directly have a look at
what archivist has manipulated when writing unit tests.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -4,6 +4,15 @@ declare type VaultOperation = 'add' | 'set' | 'del' | 'touch';
 
 declare class Archivist {
     /**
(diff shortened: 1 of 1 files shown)


ronkorving:
TBH, I would rather have a method that exposes it, rather than making something that behaves privately stick out like that. Eg:

```js
Archivist.prototype.getValueFromCache = function (topic, index) {
	var trueName = createTrueName(index || {}, topic);
	return this.loaded[trueName];
};
```

stelcheck:
In the end, it appears to be much simpler to simply make archivist calls during tests to fetch back the data, instead of trying to read this. Closing.

PR #65, opened 2017-06-09
Consul SD engine implementation

christopherobin:
This adds Consul support for MAGE, it uses the KV store to store data since the service catalog from Consul does not support storing metadata.

### DIFF
--- a/lib/serviceDiscovery/engines/consul/Readme.md
+++ b/lib/serviceDiscovery/engines/consul/Readme.md
@@ -1,0 +1,47 @@
+## Consul discovery engine
+
+Consul is a Service Discovery and distributed KV store created by HashiCorp. 
+
+### Requirements
+
+You need to install the consul module in your project:
+
+```bash
+npm install --save consul@'^0.29.0'
+```
+
+### Limitations
+
+This discovery engine needs to know on which interface your service is exposed, we might support
+auto-detection at a later time but for now if you cannot satisfy this need we recommend that you
+use a different service.
+
+### Configuration
+
+In the `server.serviceDiscovery.options` object the following options exists:
+ - __interface (required)__: The interface to advertise, such as `eth0` or `enp4s0`.
+ - __ttl (optional)__: The interval at which we should renew our consul session in milliseconds. If the server
+            crashes hard the advertisement will be removed after this TTL expires. Defaults to 30000.
+ - __consul (optional)__: Options to pass to the `node-consul` module, see [the documentation](https://github.com/silas/node-consul#consuloptions).
+
+_Sample configuration_:
+
+```yaml
+server:
+    serviceDiscovery:
+        engine: consul
+        options:
+            # Our advertised interface
+            interface: enp4s0
+            # Manually increase the TTL to 60s
+            ttl: 60000
+            # By default will connect to the local agent, override
+            # this here
+            consul:
+                host: consul.service.dc.consul
+```
+
+### Debugging
+
+Depending on the configuration of the consul server an UI should be available, at the top click
+on the `KEY/VALUE` button and go inside the `mage/` folder to find discovery keys.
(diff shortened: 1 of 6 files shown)


lib/serviceDiscovery/engines/consul/Readme.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Limitations
+
+This discovery engine needs to know on which interface your service is exposed, we might support
+auto-detection at a later time but for now if you cannot satisfy this need we recommend that you
+use a different service.
+
+### Configuration
+
+In the `server.serviceDiscovery.options` object the following options exists:
```
stelcheck: It would be nice to have a sample json or yaml configuration put right above.

lib/serviceDiscovery/engines/consul/Readme.md
```diff
@@ -0,0 +1,19 @@
...
+
+Consul is a Service Discovery and distributed KV store created by HashiCorp. 
+
+### Limitations
+
+This discovery engine needs to know on which interface your service is exposed, we might support
+auto-detection at a later time but for now if you cannot satisfy this need we recommend that you
+use a different service.
+
+### Configuration
```
stelcheck: We should mention that we need to install the consul module in user-land.

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -0,0 +1,224 @@
...
+const consul = requirePeer('consul');
+
+const mage = require('../../../mage');
+const Service = require('../../service').Service;
+const ServiceNode = require('../../node').ServiceNode;
+
+const logger = mage.core.logger.context('serviceDiscovery', 'consul');
+
+// session TTL if we crash hard
+const TTL = 30000;
```
stelcheck: This is fine as a default, but should be configurable by the user.

christopherobin:
Addressed all comments

AlmirKadric:
Apart from @stelcheck comments  from a quick glance LGTM
I'm not going to get all nit picky about the smaller things considering the circumstances

The only thing I would add:
 - Have you tested on windows?
 - Have you tested on node 4.x?
 - Did you create an issue to make unit tests later?
   (i know we need to rush it in and can't be written this time around)

AlmirKadric:
woops sorry hit the close close button by accident

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -0,0 +1,225 @@
...
+	/**
+	 * * Look at the values inside the base path and detect nodes that joined and left
+	 *
+	 * @param keys
+	 * @private
+	 */
+	_updateNetwork(keys) {
+		const currentNodes = Object.keys(this._services);
+
+		for (let item of keys) {
```
ronkorving: You should be able to use const instead of let here

lib/serviceDiscovery/service.js
```diff
@@ -2,8 +2,8 @@
  * @module serviceDiscovery
```
ronkorving: Please add a `use strict` the moment you start using `const`.

christopherobin:
@AlmirKadric 

* Have you tested on windows?

I did not but __node-consul__ is a pure JS implementation (consul uses a HTTP API) so I'm confident about it working on windows

* Have you tested on node 4.x?

Yes did my dev on node 4

* Did you create an issue to make unit tests later?

Not yet, will do. Quite time consuming since we'd need to mock the consul API/library somehow

christopherobin:
@ronkorving Fixed and rebased

ronkorving:
https://github.com/mage/mage/tree/master/lib/serviceDiscovery#configuration mentions that Zookeeper "is the only alternative". That will no longer be the case. That file also lists configuration per engine, could you add consul?

christopherobin:
@ronkorving Documentation updated

PR #69, opened 2017-06-12
types: MsgServer, MsgStream, MMRP

stelcheck:
Type information required for user-land usage of the msgServer
API.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -400,6 +400,256 @@ declare class Logger {
     emergency: Log;
 }
 
+/**
+ * There is currently only 1 flag, namely ‘TRACK_ROUTE’.
+ *
+ * When this flag is active, the returnRoute must be kept
+ * around in the envelope as it travels across the network.
+ */
+declare type MmrpEnvelopeFlag = 'NONE' | 'TRACK_ROUTE' | int
+
+/**
+ * Data which may be put into an MMRP envelope
+ */
+declare type MmrpEnvelopeMessage = string | Buffer
+
+/**
+ *
+ */
+declare type MmrpEnvelopeRoute = string[]
+
+/**
+ * MMRP Envelopes are used to encapsulate data to
+ * be sent between MAGE nodes.
+ *
+ * @class MmrpEnvelope
+ */
+declare class MmrpEnvelope {
+    constructor(eventName: string, messages: MmrpEnvelopeMessage[], route: MmrpEnvelopeRoute, returnRoute: MmrpEnvelopeRoute, flags: MmrpEnvelopeFlag | MmrpEnvelopeFlag[]);
+
+    /**
+     * Sets the message part(s) of the envelope
+     *
+     * The event name is what one will listen to, but prefixed
+     * with 'delivery.'. Example:
+     *
+     * ```javascript
+     * mmrpNode.on('delivery.MyEventName', (...args) => console.log(...args))
+     * ```
+     *
+     * @param {string} eventName
+     * @param {MmrpEnvelopeMessage} message
+     * @param {Function} callback
+     *
+     * @memberof MmrpEnvelope
+     */
+    setMessage(eventName: string, message: MmrpEnvelopeMessage): void;
+
+    /**
+     * Appends a message part to the envelope
+     *
+     * @param {MmrpEnvelopeMessage} message
+     *
+     * @memberof MmrpEnvelope
+     */
+    addMessage(message: MmrpEnvelopeMessage): void;
+
+    /**
+     * Sets the route that the envelope needs to take to reach its destination
+     *
+     * @param {MmrpEnvelopeRoute} route
+     *
+     * @memberof MmrpEnvelope
+     */
+    setRoute(route: MmrpEnvelopeRoute): void;
+
+    /**
+     * Prepends a route to the existing envelope route
+     *
+     * @param {MmrpEnvelopeRoute} route
+     *
+     * @memberof MmrpEnvelope
+     */
+    injectRoute(route: MmrpEnvelopeRoute): void;
+
+    /**
+     * Removes a route part from the start of the route, if the value matches,
+     * or empties the entire route if the given identity is undefined
+     *
+     * @param {string} [identity]
+     *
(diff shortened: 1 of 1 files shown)


types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
 }
 
+/**
+ * MMRP Envelopes are used to encapsulate data to
+ * be sent between MAGE nodes.
+ *
+ * @class MmrpEnvelope
+ */
+declare class MmrpEnvelope {
+    constructor(eventName: string, messages: string | Buffer | string[] | Buffer[], route: string[], returnRoute: string[], flags: any);
```
ronkorving: flags are: string, int, string[] or int[]

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+    /**
+     *
+     *
+     * @param {string} uri
+     * @param {string} clusterId
+     * @param {any} cb
+     *
+     * @memberof MmrpEnvelope
+     */
+    setMessage(uri: string, clusterId: string, callback: Function): void;
```
ronkorving: signature is (type, message)

Something went wrong here.

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+ * MMRP Envelopes are used to encapsulate data to
+ * be sent between MAGE nodes.
+ *
+ * @class MmrpEnvelope
+ */
+declare class MmrpEnvelope {
+    constructor(eventName: string, messages: string | Buffer | string[] | Buffer[], route: string[], returnRoute: string[], flags: any);
+
+    /**
+     *
```
ronkorving: > Sets the message part(s) of the envelope

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     * @param {string} uri
+     * @param {string} clusterId
+     * @param {any} cb
+     *
+     * @memberof MmrpEnvelope
+     */
+    setMessage(uri: string, clusterId: string, callback: Function): void;
+
+    /**
+     *
```
ronkorving: > Appends a message part to the envelope

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @param {(string | Buffer)} message
+     *
+     * @memberof MmrpEnvelope
+     */
+    addMessage(message: string | Buffer): void;
+
+    /**
+     *
```
ronkorving: > Sets the route that the envelope needs to take to reach its destination

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @param {(string | string[])} route
+     *
+     * @memberof MmrpEnvelope
+     */
+    setRoute(route: string | string[]): void;
+
+    /**
+     *
```
ronkorving: > Prepends a route to the existing envelope route

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @param {string} route
```
ronkorving: May also be an array of strings

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @memberof MmrpEnvelope
+     */
+    injectRoute(route: string): void;
+
+    /**
+     *
+     *
+     * @param {*} identity
```
ronkorving: Must be string or undefined (to empty the route)

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     *
+     * @param {string} route
+     *
+     * @memberof MmrpEnvelope
+     */
+    injectRoute(route: string): void;
+
+    /**
+     *
```
ronkorving: > Removes a route part from the start of the route, if the value matches. Or empties the entire route if the given identity is undefined.

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @param {*} identity
+     *
+     * @memberof MmrpEnvelope
+     */
+    consumeRoute(identity: any): void;
+
+    /**
+     *
```
ronkorving: > Returns true if the route has not been fully consumed, false otherwise.

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @returns {boolean}
+     *
+     * @memberof MmrpEnvelope
+     */
+    routeRemains(): boolean;
+
+    /**
+     *
```
ronkorving: > Returns the final address in the route.

types.d.ts
```diff
@@ -400,6 +400,225 @@ declare class Logger {
...
+     *
+     * @returns {string}
+     *
+     * @memberof MmrpEnvelope
+     */
+    getFinalDestination(): string;
+
+    /**
+     *
```
ronkorving: Returns true if this envelope tracks and contains the sender's route, false otherwise.

types.d.ts
stelcheck: Where may I find a description of their meaning?

stelcheck: NM, found it in the API doc; however, what would happen if I specified both NONE and TRACK_ROUTE in an array?

ronkorving: they're bitwise ORed, and NONE is 0, so it would simply be TRACK_ROUTE.

types.d.ts
stelcheck: According to the code, it can only be an array of string, it seems.

types.d.ts
```diff
@@ -400,6 +400,256 @@ declare class Logger {
...
+
+    /**
+     * Set a flag value
+     *
+     * @param {*} flag
+     * @returns {boolean}
+     *
+     * @memberof MmrpEnvelope
+     */
+    setFlag(flag: any): boolean;
```
ronkorving: string or int for the flag param?

ronkorving:
LGTM, you can merge after CI 👍

PR #70, opened 2017-06-14
link-require -> app-module-path, mocha.opts

stelcheck:
Moving mocha options to ./test/mocha.opts

Replacing link-require with app-module-path, since
link-require breaks on the new NPM.


ronkorving:
Just a thought that crossed my mind the other day. These solutions, don't they render it completely impossible for the end-user to use this for their own project, with MAGE hyjacking the "lib" alias?

stelcheck:
No, because of scoping.

PR #73, opened 2017-06-21
Fix several issues with consul discovery

christopherobin:
Sorry about those, it should have been part of my testing procedure.

### Fixes

* Ignored error during session creation
* Uncaught error in the watcher if consul goes down (or network goes bad)
* Added re-announce mechanic to try and re-create the session if the server
  comes back up

### Console output

image

### DIFF
--- a/lib/serviceDiscovery/engines/consul/index.js
+++ b/lib/serviceDiscovery/engines/consul/index.js
@@ -71,7 +71,12 @@ class ConsulService extends Service {
 				this._sessionId = id.ID;
 				this._keepaliveTimer = setTimeout(this._keepalive.bind(this, this._sessionId), this.ttl);
 
-				logger.debug(`acquired session ID ${this._sessionId}`);
+				logger.debug(`Acquired session ID ${this._sessionId}`);
+			}
+
+			// transform the error to something more friendly
+			if (err) {
+				logger.error.data(err).log('Couldn\'t connect to Consul');
 			}
 
 			cb(err, this._sessionId);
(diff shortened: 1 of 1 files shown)


stelcheck:
@ronkorving this is rather strange; travis succeeded, but not appveyor https://ci.appveyor.com/project/mage/mage/build/1.0.173/job/11eheq3p2jj4vfgg#L633

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -116,13 +124,19 @@ class ConsulService extends Service {
 			this._updateNetwork(data || []);
 		});
 
+		this._watch.on('error', (error) => {
+			logger.error.data(error).log('error while running discovery process');
```
stelcheck: Is a retry/recovery logic provided somewhere else by any chance?

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -116,13 +124,19 @@ class ConsulService extends Service {
...
+		});
+
 		logger.debug('discovery started');
 	}
 
 	/**
 	 * Stop discovery and announcements
 	 */
 	close() {
+		this._announces = [];
```
stelcheck: Might be safer to blow up the content of the array instead of overwriting it, in case someone makes a reference to it.

lib/serviceDiscovery/engines/consul/index.js
christopherobin: Yes, with exponential backoff

lib/serviceDiscovery/engines/consul/index.js
ronkorving: If people wanna abuse private members, they do so at their own risk. We shouldn't have to jump through hoops just in case, imho.

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -125,17 +132,15 @@ class ConsulService extends Service {
...
-
-		logger.debug('discovery started');
 	}
 
 	/**
 	 * Stop discovery and announcements
 	 */
 	close() {
-		this._announces = [];
+		this._announces.splice(0, this._announces.length);
```
ronkorving: If you guys really wanna go there, an easier (and much faster) way to empty out an array:

```js
this._announces.length = 0;
```

PR #76, opened 2017-06-22
graceful worker exit from worker

stelcheck:
In some cases, a developer might want to be able to
shut down a worker and have it respawned by the master
process. By also taking into consideration
the error code, we can now allow for this behavior.


ronkorving:
This only changes logging, no actual behavior. Is that intended?

stelcheck:
Yes. The process will get reloaded, so that is not a problem; the issue is that there were no ways to trigger a reload from the worker process without triggering a rather scary emergency log.

ronkorving:
Got it, then I'm happy with this change 👍

PR #77, opened 2017-06-26
Have the start command output stderr when failing

christopherobin:
Currently when the start command fails the use will be greeted by this quite not helpful message:

```haskell
$ npm run start

> my-awesome-project@1.0.0 start /my/awesome/project
> echo             'Start MAGE as a daemon'                 && mage start

Start MAGE as a daemon
Starting "/path/to/node /my/awesome/project/node_modules/.bin/mage"... (pid: 999)
Failed.
```

This changes has the deamonizer keep `stderr` on the child as a pipe, and if an error occur will pipe it to the current processes `stderr`:

```haskell
$ npm run start

> my-awesome-project@1.0.0 start /my/awesome/project
> echo             'Start MAGE as a daemon'                 && mage start

Start MAGE as a daemon
Starting "/path/to/node /my/awesome/project/node_modules/.bin/mage"... (pid: 1000)
Failed.

m-1000 - 13:37:32.171     <debug> [MAGE archivist] Creating persistent vaults ["volatileVault","userVault"]
... snip ...
m-1000 - 13:37:32.177 <emergency> [MAGE] Error during MAGE setup: No valid IPv4 found on interface invalid
  data: {
    "file": "index.js",
    "line": 39,
    "offset": 10,
    "type": "Error",
    "stack": [ "... snip ..." ]
  }
m-1000 - 13:37:32.177     <debug> [MAGE] Shutting down Mage...
m-1000 - 13:37:32.179     <debug> [MAGE archivist] Closing all vaults...
m-1000 - 13:37:32.181     <debug> [MAGE] Destroying all log writers

### DIFF
--- a/lib/daemon/index.js
+++ b/lib/daemon/index.js
@@ -105,7 +105,7 @@ function App(runstate, pid) {
 }
 
 
-function sendSignal(pid, signal, timeoutMsec, cb) {
+function sendSignal(app, signal, timeoutMsec, cb) {
 	var timeout, sigSuccessHandler, sigProgressHandler, sigFailedHandler;
 
 	function done(exitCode) {
@@ -144,6 +144,7 @@ function sendSignal(pid, signal, timeoutMsec, cb) {
 
 	// Request failed
 	sigFailedHandler = function () {
+		app.stderr.pipe(process.stderr);
 		err(chalk.red.bold('Failed.'));
 		done(EXIT_FAILED);
 	};
@@ -156,9 +157,9 @@ function sendSignal(pid, signal, timeoutMsec, cb) {
 	process.on(RESP_SUCCESS, sigSuccessHandler);
 	process.on(RESP_FAILED, sigFailedHandler);
 
-	if (pid && signal) {
+	if (app && app.pid && signal) {
 		try {
-			process.kill(pid, signal);
+			process.kill(app.pid, signal);
 		} catch (e) {
 			err(chalk.yellow.bold('Process has gone away'));
 			done(EXIT_FAILED);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by christopherobin** [comment]
Actually ran into issues with the child dying since the `stderr` pipe is owned by the daemonizer and is destroyed once it exists, causing the child to die on next write. We need to find a way to tell the child to stop using `stderr`, or redirect it to `/dev/null`.

File: lib/daemon/index.js [hunk 1]
```diff
@@ -188,18 +189,19 @@ function start(app, cb) {
...
 
 	out(chalk.green.bold('Starting "' + appName + ' ' + args.join(' ') + '"... (pid: ' + child.pid + ')'));
 
 	// wait for a signal from the child process before we continue
 
-	sendSignal(null, null, timeouts.start, function (exitCode) {
+	sendSignal(app, null, timeouts.start, function (exitCode) {
 		// don't count the spawned app in the reference count of the daemonizer
 
 		child.unref();
```
stelcheck: I think adding 

```javascript
child.stderr.unpipe(app.stderr)
```
right here would solve the issue you have mentioned.

ronkorving: May well indeed. Worth a try.

christopherobin: Does not seem like it's enough, game still dies violently soon after starting, tried to enable the syslog writer on top of the terminal one and I can't get any error message

stelcheck:
Try manually piping instead of piping through the spawn option, then
unpiping, just to be sure. I wouldn't be surprised if spawn used a
different reference to do the initial piping, this will make sure to bypass
any black magic it might be doing.

ronkorving:
I think stderr is not a simple stream, and this mechanism isn't using stream pipes. You may get lucky trying to pipe it manually, but I doubt it'll work :(

stelcheck:
@christopherobin did you get the chance to try what we discussed last time?

A quick recap of discussion: when daemonizing, we could very well force-start/configure (would assume a >=notice config would be sufficient for those cases) the terminal logger (in case it is disabled by config), and force-stop it once we have confirmed that startup has occurred successfully.

christopherobin:
Not yet, I'll take a look at trying it today in the afternoon

stelcheck:
https://github.com/stelcheck/mage/commit/7357938abc40fa4996f3ad5b57cb5d4c2bb12e6e Should solve your issue.

stelcheck:
@christopherobin do you think you could apply the suggested fix this week? We will be having a release some time soon, and I think this could be a worthy addition to it.

christopherobin:
Rebased with @stelcheck's changes, works fine for me

PR #81, opened 2017-06-30
Let consul default on the first interface when not configured

christopherobin:
The restrictive nature of the consul module right now makes it hard on developers since based on the environment the interface might be different.

This fixes it by making it possible to omit the consul interface, in this configuration it will pick the first available interface and emit a warning instead of crashing hard.

### DIFF
--- a/lib/serviceDiscovery/engines/consul/index.js
+++ b/lib/serviceDiscovery/engines/consul/index.js
@@ -24,7 +24,25 @@ class ConsulService extends Service {
 		super();
 
 		if (!options.interface) {
-			throw new Error('Missing "interface" option for consul module');
+			// find the first interface with a valid IPv4
+			let validIface;
+
+			const interfaces = os.networkInterfaces();
+			for (const iface of Object.keys(interfaces)) {
+				if (interfaces[iface].filter(v => v.family === 'IPv4').length) {
+					validIface = iface;
+					break;
+				}
+			}
+
+			if (!validIface) {
+				throw new Error('No valid interface found on this device');
+			}
+
+			// take the first one
+			options.interface = validIface;
+
+			logger.warning(`No interface defined for consul, defaulting to ${options.interface}`);
 		}
 
 		this.name = name;
--- a/lib/serviceDiscovery/engines/consul/Readme.md
+++ b/lib/serviceDiscovery/engines/consul/Readme.md
@@ -19,7 +19,7 @@ use a different service.
 ### Configuration
 
 In the `server.serviceDiscovery.options` object the following options exists:
- - __interface (required)__: The interface to advertise, such as `eth0` or `enp4s0`.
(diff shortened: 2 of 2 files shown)


lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -24,7 +24,29 @@ class ConsulService extends Service {
 		super();
 
 		if (!options.interface) {
-			throw new Error('Missing "interface" option for consul module');
+			// find the first interface with a valid IPv4
+			let validIface;
+
+			const interfaces = os.networkInterfaces();
+			for (let iface in interfaces) {
+				if (!interfaces.hasOwnProperty(iface)) {
```
ronkorving: MAGE blew up in logging because more and more objects that Node returns don't have a prototype, and thus don't have `.hasOwnProperty`. You may wanna use `Object.keys()` and iterate with a regular for-loop.

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -24,7 +24,25 @@ class ConsulService extends Service {
...
+			}
+
+			if (!validIface) {
+				throw new Error('No valid interface found on this device');
+			}
+
+			// take the first one
+			options.interface = validIface;
+
+			logger.warning(`No interface defined for consul, defaulting to ${options.interface}`);
```
ronkorving: You can just use `validIface` here. Not a blocker.

ronkorving:
What's the reasoning behind "the first"? As far as Node.js is concerned, I don't think there is a natural order of interfaces, since they're returned as a key/value object, not as an array.

christopherobin:
I'm just taking the first in the array, not on the machine, though in my testing with a machine that had 10+ interfaces, it would consistently match `lo` so node might be doing some magic there?

ronkorving:
The array that you create, right? Node is creating an object that you traverse. AFAIK JavaScript does not guarantee property order in objects, although of course the V8 runtime might be doing that internally. That is, `Object.keys(obj)` may consistently return the keys in the order in which they were added to the object. That's not something I like to depend on though.

ronkorving:
Would it not be safer to find the first interface that has `internal: true` on its IPv4 address?

christopherobin:
There not really any security issue there, we are just announcing an IP to consul and any IPv4 is perfectly fine, I could probably also just send 127.0.0.1.

christopherobin:
After all we are dealing with a scenario where the user didn't configure their consul properly. Taking the first internal will not always work since for example bridges and vlans are marked as internal in some cases. If we'd want the proper IP we'd need to find the default route of the machine but node.js doesn't provide this information.

ronkorving:
Sorry, when I meant "safe", I really just meant "a safe bet" (not security).

ronkorving:
Discussed on chat, we should promote the use of configuration, which can be aided by environment variables if it's different for all users.

PR #84, opened 2017-07-03
Fixes command center logging when ACL test fails

ronkorving:
This log entry was poorly formatted. It was:

```
Access level required is "admin
" but user access level is "user
```

Note the missing quotation mark at the end, and the misplaced quotation mark after `admin`.
Output is now:

```
Access level required is "admin",
but user access level is "user"
```

Keep in mind that every call to `.details()` doesn't just concatenate, it also creates a newline.


stelcheck:
I believe this was the intended behavior, hence why there were no spaces after the comma being removed by this PR. 

Other than that, I am not entirely sure what the point of this change actually is; neither the commit message (which has no body) nor the PR body explicitly mention what is being fixed here.

ronkorving:
Let me update it with output samples.

ronkorving:
@stelcheck PR description updated.

PR #92, opened 2017-07-13
Suppress two vulnerable packages: growl, brace-expansion

ronkorving:
While they are unsafe, they are only used in development.

**growl**: We need to wait for a growl release and for mocha to get updated.
**brace-expansion**: This has been fixed, but sqlite3 needs a release to update its dependencies.


AlmirKadric:
LGTM 👍

PR #95, opened 2017-07-14
bin/mage: deal with early errors

### LINKED ISSUE #89: mage binary running boot on non-mage instance results in quiet failure
Currently if you accidentally forget to module.export mage from your `lib/index.js` or accidentally configure your main file to be something other than the `lib/index.js` file, mage will quietly crash. Rather than quietly crashing our uncaught exception handler should write to stdout if a logger hasn't yet been setup yet.

stelcheck:
- Deal with early errors occuring in the app's entrypoint
  - Deal with app not exporting mage back

Fixes #89

### DIFF
--- a/bin/mage
+++ b/bin/mage
@@ -1,4 +1,47 @@
 #!/usr/bin/env node
 
-var app = require(process.cwd());
+const path = require('path');
+const cwd = process.cwd();
+
+let app;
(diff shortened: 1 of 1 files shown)


stelcheck:
@AlmirKadric updates?

PR #97, opened 2017-07-20
mdns2 -> mdns

### LINKED ISSUE #63: mDNS
I don't mind that we support mDNS, and I'm in favor of maintaining the support for now, but we enforce the `mdns2` peer dependency. A long time ago, `mdns` got outdated, and we updated it and labeled it `mdns2`. However, things have (thankfully) reversed and `mdns` is again pretty well maintained. We should change the peer dependency requirement.

stelcheck:
mdns2 does not appear to work properly on newer versions of Node (tested
on Node 7). This reverts back to the latest mdns version, which appears
to be working just fine.

Fixes #63


ronkorving:
Yeah, the mdns module got attention again, making mdns2 obsolete.

stelcheck:
It's not only that, it doesn't seem to compile on latest node versions.

ronkorving:
Not surprising. Anything that doesn't receive NAN updates at some point will stop compiling on newer major node releases.

PR #98, opened 2017-07-21
Module teardown

stelcheck:
Allow for modules to specify a teardown function, which
essentially does the opposite of setup (runs at shutdown time).

### DIFF
--- a/docs-sources/includes/_Modules.md
+++ b/docs-sources/includes/_Modules.md
@@ -17,7 +17,7 @@ functionality for a certain subject (users, ranking, shop, etc),
 and expose API (called user commands) that are accessible through
 the different client SDKs.
 
-## Module setup
+## Module setup and teardown
 
 > lib/modules/players/index.js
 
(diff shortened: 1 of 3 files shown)


docs-sources/includes/_Modules.md
```diff
@@ -26,15 +26,28 @@ exports.setup = function (state, callback) {
...
+The setup function will run when MAGE boots up allowing your module
 to prepare itself, for example by loading vital information from a data store into memory.
 This function is optional, so if you do not have a setup phase, you don't need to add it.
 
-Note that by default, modules will have 5000 milliseconds to complete setup; should you need
-longer than that, you will need to set `exports.setupTimeout` to the value of your choice.
+Teardown functions in a similar way; you may want to store to database
+some data you have been keeping in memory, or send some notifications to your users.
+
+Note that by default, each hooks will have 5000 milliseconds to complete; should you need
```
ronkorving: "each hook"

stelcheck: Fixed.

lib/tasks/serve.js
```diff
@@ -172,6 +180,7 @@ exports.shutdown = function (mage, options, cb) {
 	}
 
 	async.series([
+		teardownModules,
 		closeServers,
```
ronkorving: I thought the intention was to close servers first, to avoid requests coming in and ending up in our modules. Did your reasoning on this change?

stelcheck: `closeServers` would close both mmrp and http. I would need to split that phase into two, and teardown in between; is that OK?

ronkorving: Works for me 👍

stelcheck: OK, fixed

PR #99, opened 2017-07-21
log data: log anything, anywhere

### LINKED ISSUE #96: logger data: passing an array outputs incorrect data
Given:

```javscript
const logger = mage.logger.context('send')
logger.debug.data(presents).log('presents')
logger.debug(presents[0].userId)
```

The output is:

```plaintext
w-3319 - 01:12:50.622     <debug> [app send] send presents
  data: [
    []
  ]
w-3319 - 01:12:50.632     <debug> [app send] aa5249c7-8c84-45aa-abe1-b3450b01770b
```

I expect that other non-object literals would also cause similar issues (although I have not tested).

I see only two options to solve this issue:

  1. Warn or error out on non-objects; this is not too ideal, since some logs may seldom pop up (so you would only figure out that there is an issue in the worst cases)
  2. Assign a label to it automatically (although I am not sure what that label should be)

@AlmirKadric @ronkorving let me know what you guys think

stelcheck:
- logEntry.data() can now take any data
- label remains optional (default label will be data)
- Data is kept in an array instead of in an object

Now looks like:

screenshot_2017-07-21_16-02-42

screenshot_2017-07-21_16-01-48

screenshot_2017-07-21_16-02-11

screenshot_2017-07-21_16-05-13

Fixes #96

### DIFF
--- a/lib/loggingService/LogEntry.js
+++ b/lib/loggingService/LogEntry.js
@@ -9,8 +9,63 @@ var rootPath = process.cwd();
 var IncomingMessage = require('http').IncomingMessage;
 var ServerResponse = require('http').ServerResponse;
 
+function serializeError(error) {
+	// the stack trace and other error information will be moved into the data object
+
+	// turn the stack string into an array, stripped from its noisy whitespace
+
+	var stack = error.stack && error.stack.split(/\s*\n\s*at\s+/);
+	if (!stack || stack.length <= 1) {
+		return error;
+	}
+
+	// remove the first line: "Error: name", leaving only stack frames
+
+	stack.shift();
+
+	// assign all the data to the data object (custom errors may have some)
+	var data = Object.assign({
+		code: error.code,
+		type: error.name,
+		message: error.message || undefined
+	}, error);
+
+	// reformat the stack
+
+	var locationFound = false;
+
+	for (var i = 0; i < stack.length; i += 1) {
+		// make the file path relative to the project
+
+		stack[i] = stack[i].replace(rootPath + '/', '');
+
+		// If the error did not originate in a native function (like JSON.parse), the top stack
+		// frame should carry file, line and character offset information. While we're looking
+		// at a native function, we move on to the next stack frame to get us closest to the
+		// source of the error in user-land code.
+
+		if (!locationFound) {
+			var m = stack[i].match(/([\w\.]+):([0-9]+):([0-9]+)/);
+			if (m) {
+				locationFound = true;
+
+				data.file = m[1];
+				data.line = parseInt(m[2], 10) || 0;
+				data.offset = parseInt(m[3], 10) || 0;
+			}
+		}
+	}
+
+	data.stack = stack;
+
+	return data;
+}
 
 function serializeAny(obj, refList) {
+	if (obj && obj.value instanceof Error) {
+		return { label: obj.label, value: serializeError(obj.value) };
+	}
+
 	var out, i;
 
 	refList = refList || [];
@@ -110,68 +165,13 @@ function LogEntry(channel) {
 }
 
 
-LogEntry.prototype.registerErrorDetails = function (error) {
-	// the stack trace and other error information will be moved into the data object
-
-	// turn the stack string into an array, stripped from its noisy whitespace
-
-	var stack = error.stack && error.stack.split(/\s*\n\s*at\s+/);
-	if (!stack || stack.length <= 1) {
-		return;
-	}
-
-	// remove the first line: "Error: name", leaving only stack frames
-
-	stack.shift();
-
-	this.data = this.data || {};
-
-	// assign all the data to the data object (custom errors may have some)
-	Object.assign(this.data, error);
-
-	// Remove message, we already display it in the log text
-	delete this.data.message;
-
-	// reformat the stack
-
-	var locationFound = false;
-
-	for (var i = 0; i < stack.length; i += 1) {
-		// make the file path relative to the project
-
-		stack[i] = stack[i].replace(rootPath + '/', '');
-
-		// If the error did not originate in a native function (like JSON.parse), the top stack
-		// frame should carry file, line and character offset information. While we're looking
-		// at a native function, we move on to the next stack frame to get us closest to the
-		// source of the error in user-land code.
-
-		if (!locationFound) {
-			var m = stack[i].match(/([\w\.]+):([0-9]+):([0-9]+)/);
-			if (m) {
-				locationFound = true;
-
-				this.data.file = m[1];
-				this.data.line = parseInt(m[2], 10) || 0;
-				this.data.offset = parseInt(m[3], 10) || 0;
-			}
-		}
-	}
-
-	if (error.code) {
-		this.data.code = error.code;
-	}
-
-	this.data.type = error.name;
-	this.data.stack = stack;
-};
-
-
 LogEntry.prototype.serializeArgument = function (arg) {
 	if (arg instanceof Error) {
-		this.registerErrorDetails(arg);
-
+		const error = serializeError(arg);
+		delete error.message;
 		arg = arg.message;
+
+		this.addData([{ label: 'error', value: error }]);
 	}
 
 	if (arg === undefined) {
@@ -247,16 +247,12 @@ LogEntry.prototype.addDetails = function (args) {
 
 
 LogEntry.prototype.addData = function (data) {
-	data = serializeAny(data);
+	if (!this.data) {
+		this.data = [];
+	}
 
-	if (this.data) {
-		for (var key in data) {
-			if (data.hasOwnProperty(key)) {
-				this.data[key] = data[key];
-			}
-		}
-	} else {
-		this.data = data;
+	for (const entry of data) {
+		this.data.push(serializeAny(entry));
 	}
 };
 
--- a/lib/loggingService/writers/file.js
+++ b/lib/loggingService/writers/file.js
@@ -200,7 +200,9 @@ FileWriter.prototype.channelFunctionGenerator = function (channelName) {
 		// additional info, formatted (JSON)
 
 		if (entry.data) {
-			msg += '\n' + JSON.stringify(entry.data, null, jsonIndent);
+			for (const dataEntry of entry.data) {
+				msg += `\n${dataEntry.label}: ` + JSON.stringify(dataEntry.value, null, jsonIndent);
+			}
 		}
 
 		// We use the ISOString format for the date, it as written as an UTC date using the format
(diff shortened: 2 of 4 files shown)


stelcheck:
Updated the description with some screenshots.

For some reason, this breaks error logging; error objects are not properly parsed any longer.

stelcheck:
Fixed error logging. Now you can also log data alongside an error object.

```javascript
require('mage').logger.error(new Error('test one'))
```

screenshot_2017-07-21_19-10-31

Behaviour is the same as usual, except that the label is automatically set to `error`

```javascript
    require('mage').logger.error
        .data([1, 2, 3])
        .log(new Error('test one'))
```

screenshot_2017-07-21_19-13-18

Same as before, but now you can also log additional data alongside the error.

```javascript
    require('mage').logger.error
      .data([ 1, 2, 3])
      .data(new Error('test two'))
      .log('Failed to do something really important')
```

screenshot_2017-07-21_19-07-29

Here, we set the log text to something custom; in this case, the error object being logged keep the `message` attribute (instead of removing it like in previous cases).

stelcheck:
I need for us to confirm that this PR does not break the other log writers not touched by this PR. Would appreciate some help with that, if possible.

lib/loggingService/writers/terminal.js
```diff
@@ -159,7 +159,9 @@ TerminalWriter.prototype.channelFunctionGenerator = function (channelName) {
 		// additional info, formatted (JSON)
 
 		if (entry.data) {
-			msg += '\ndata: ' + JSON.stringify(entry.data, null, jsonIndent);
+			for (const dataEntry of entry.data) {
+				msg += `\n${dataEntry.label}: ` + JSON.stringify(dataEntry.value, null, jsonIndent);
```
AlmirKadric: mixture of template and JSON.stringify
eslint should complaint about this

lib/loggingService/writers/file.js
```diff
@@ -200,7 +200,9 @@ FileWriter.prototype.channelFunctionGenerator = function (channelName) {
 		// additional info, formatted (JSON)
 
 		if (entry.data) {
-			msg += '\n' + JSON.stringify(entry.data, null, jsonIndent);
+			for (const dataEntry of entry.data) {
+				msg += `\n${dataEntry.label}: ` + JSON.stringify(dataEntry.value, null, jsonIndent);
```
AlmirKadric: as below

lib/loggingService/writers/terminal.js
stelcheck: CI does not, so I don7t think that is an issue cc @ronkorving

lib/loggingService/writers/file.js
```diff
@@ -200,7 +200,9 @@ FileWriter.prototype.channelFunctionGenerator = function (channelName) {
 		// additional info, formatted (JSON)
 
 		if (entry.data) {
-			msg += '\n' + JSON.stringify(entry.data, null, jsonIndent);
+			for (const dataEntry of entry.data) {
```
stelcheck: https://github.com/mage/mage/pull/99/files/9ef0bd6deb77db99c4477af5be2a37d9cb222558#diff-c3caf28db5144d0bfc663414f3eeaf12R250

This should always be an array or undefined.

lib/loggingService/writers/terminal.js
```diff
@@ -159,7 +159,9 @@ TerminalWriter.prototype.channelFunctionGenerator = function (channelName) {
 		// additional info, formatted (JSON)
 
 		if (entry.data) {
-			msg += '\ndata: ' + JSON.stringify(entry.data, null, jsonIndent);
+			for (const dataEntry of entry.data) {
```
stelcheck: Same here, `entry.data` is an array or null.

lib/loggingService/LogEntry.js
```diff
@@ -247,17 +247,11 @@ LogEntry.prototype.addDetails = function (args) {
...
-	if (this.data) {
-		for (var key in data) {
-			if (data.hasOwnProperty(key)) {
-				this.data[key] = data[key];
-			}
-		}
-	} else {
-		this.data = data;
+	if (!this.data) {
+		this.data = [];
```
stelcheck: We push data in an array of objects, and the value is set as an attribute of that object. Data should be anything you want, including undefined.

lib/loggingService/LogEntry.js
```diff
@@ -110,68 +165,13 @@ function LogEntry(channel) {
...
-
-
 LogEntry.prototype.serializeArgument = function (arg) {
 	if (arg instanceof Error) {
-		this.registerErrorDetails(arg);
-
+		const error = arg;
 		arg = arg.message;
+
+		delete error.message;
```
stelcheck: See https://github.com/mage/mage/pull/99/files/9ef0bd6deb77db99c4477af5be2a37d9cb222558#diff-c3caf28db5144d0bfc663414f3eeaf12L133 cc @ronkorving 

It makes sense to me that we shouldn't tamper with the original error, however, so I won't mind fixing this; just want to make sure we don't perhaps already copy upstream.

lib/loggingService/writers/file.js
AlmirKadric: understood 👍

lib/loggingService/writers/terminal.js
AlmirKadric: understood 👍

lib/loggingService/LogEntry.js
```diff
@@ -247,17 +247,11 @@ LogEntry.prototype.addDetails = function (args) {
...
-				this.data[key] = data[key];
-			}
-		}
-	} else {
-		this.data = data;
+	if (!this.data) {
+		this.data = [];
 	}
+
+	this.data = data.map((entry) => serializeAny(entry)).concat(this.data);
```
AlmirKadric: hmmm this line works?
the arrow function isn't returning the value from serializeAny according to what is written here.
also you're overwriting this.data and forcing a new array to be constructed every time
why not simplify into the following?
```
this.data.push(...data.map(serializeAny))
```

stelcheck: Arrow functions have automatic return when written like this.

AlmirKadric: ah i see, did not know that 勉強になりました ^_^
what about simplifying the line using `...`?
Though this will mean mage will have stricter node version requirements.
@ronkorving thoughts?

stelcheck: Merh, I'll optimize this bit. No need for double-iteration anyway

AlmirKadric:
In regards to the log writers, apart from the client side rendering of the websocket writer (since data format has changed), everything else should work fine from what I have seen. I can't seem to find the websocket rendering code, it used to be inside the MAGE dashboard system, but not sure if we scrapped that code along with the dashboard itself. @ronkorving

lib/loggingService/LogEntry.js
```diff
@@ -9,8 +9,63 @@ var rootPath = process.cwd();
...
+
+	// remove the first line: "Error: name", leaving only stack frames
+
+	stack.shift();
+
+	// assign all the data to the data object (custom errors may have some)
+	var data = Object.assign({
+		code: error.code,
+		type: error.name,
+		message: error.message || undefined
```
ronkorving: You might as well assign to an empty object then, since you're copying over all properties.
However, if you want to keep it this way (to ensure these properties *always* exist), I would suggest removing `code`. It's not a standard (a convention at best).

lib/loggingService/LogEntry.js
```diff
@@ -9,8 +9,63 @@ var rootPath = process.cwd();
...
+		return error;
+	}
+
+	// remove the first line: "Error: name", leaving only stack frames
+
+	stack.shift();
+
+	// assign all the data to the data object (custom errors may have some)
+	var data = Object.assign({
+		code: error.code,
```
ronkorving: You might as well assign to an empty object then, since you're copying over all properties.
However, if you want to keep it this way (to ensure these properties *always* exist), I would suggest removing `code`. It's not a standard (a convention at best).

lib/loggingService/LogEntry.js
```diff
@@ -9,8 +9,63 @@ var rootPath = process.cwd();
...
+				locationFound = true;
+
+				data.file = m[1];
+				data.line = parseInt(m[2], 10) || 0;
+				data.offset = parseInt(m[3], 10) || 0;
+			}
+		}
+	}
+
+	data.stack = stack;
```
ronkorving: The `stack` array can be assigned in the `Object.assign` above.

lib/loggingService/LogEntry.js
ronkorving: I'm not a fan of making a deep copy of error objects. The `delete` could happen on the return value of `serializeError`.

lib/loggingService/LogEntry.js
stelcheck: True, its not standard... but many errors even in node core libs include it. Since it was already in the previous code (I believe), I thought it would be wise to keep it.

lib/loggingService/LogEntry.js
stelcheck: While not technically guaranteed, I make sure to assign stack last because I want it to show as the last attribute when printed on the screen - I find it easier to read.

lib/loggingService/LogEntry.js
stelcheck: Fixed, please see new code.

stelcheck:
@ronkorving still waiting for an answer to what I said concerning the last two comments you have made, namely:

https://github.com/mage/mage/pull/99#discussion_r130033384
https://github.com/mage/mage/pull/99#discussion_r130033196

ronkorving:
@stelcheck Those are not blockers, just nitpicking.

ronkorving:
@stelcheck Is the Do Not Merge label still relevant?

stelcheck:
@ronkorving do not merge is in regard to this: https://github.com/mage/mage/pull/99#issuecomment-318253590

We need someone to confirm that the changes introduced here won't be breaking other log backends; I don't think it would, but just want someone else to confirm it. @AlmirKadric has reviewed the other ones (a second pair of eyes wouldn't hurt), but we still are not sure about websocket backend.

ronkorving:
I reviewed them all.

I don't believe we currently have a consumer for the WebSocket writer output, so it should be safe 👍  Maybe @christopherobin or @tokyowizard can confirm is they're using it anywhere, perhaps in some dark corner of their ops machine? ;)

christopherobin:
Never used

PR #101, opened 2017-07-24
Fix service discovery not being shutdown properly

christopherobin:
In some cases service discovery shutdown can be asynchronous (ie. `consul`), this adds a callback to the `msgServer.close` function that is passed all the way down to SD, allowing them to exit cleanly.

image

### DIFF
--- a/lib/tasks/serve.js
+++ b/lib/tasks/serve.js
@@ -157,8 +157,7 @@ exports.shutdown = function (mage, options, cb) {
 	}
 
 	function closeMsgServer(callback) {
-		mage.core.msgServer.close();
-		callback();
+		mage.core.msgServer.close(callback);
 	}
 
 	function closeVaults(callback) {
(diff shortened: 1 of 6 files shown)


lib/tasks/serve.js
```diff
@@ -145,8 +145,10 @@ exports.start = function (mage, options, cb) {
 
 exports.shutdown = function (mage, options, cb) {
 	function closeServers(callback) {
-		mage.core.msgServer.close();
-		mage.core.httpServer.close(callback);
+		async.waterfall([
```
ronkorving: Could you `async.series` instead? Waterfall has a very nasty habit (well, by design) to pass along arguments to the next function, which means that if you call `cb(null)` anywhere for example, your `next` will become `null` and the callback will be the 2nd argument you don't receive.

ronkorving: On 2nd thought, @stelcheck just did a PR (#98) that splits `closeServers` up into two, so you don't have to worry about it if you pull develop.

PR #105, opened 2017-07-26
Additional slashes in a path are cleaned up

### LINKED ISSUE #74: Superfluous slashes in an URL should be processed properly
Example:  

  * https://localhost:8080/game/player.login
  * https://localhost:8080/game//player.login
  * https://localhost:8080//game/player.login

Should all resolve to the same usercommand - at the moment, the last two ones would give you a 404.

@AlmirKadric @ronkorving

stelcheck:
Fixes #74

### DIFF
--- a/lib/httpServer/transports/http/HttpRouter.js
+++ b/lib/httpServer/transports/http/HttpRouter.js
@@ -1,3 +1,5 @@
+const resolve = require('path').posix.resolve;
+
 function HttpRouter(types) {
 	this.types = types;
 	this.exact = {};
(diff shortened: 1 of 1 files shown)


stelcheck:
One issue I see with this PR is that logs are not showing a clean path, but the received path instead.

screenshot_2017-07-26_15-14-07

Not really an issue in development, but perhaps could be an issue in production? Are there any places in our logging where we would want to log the clean path instead?

christopherobin:
I'm fine with this but at the same time I feel like the SDK should never generate this kind of URI unless we have a bug somewhere

stelcheck:
This is not for the SDKs, but for people who may want to use cURL or external tools like https://chrome.google.com/webstore/detail/advanced-rest-client/hgmloofddffdnphfgcellkdfbfbjeloo; I had cases where through some copy-pasta devs ended up adding an extra slash, and then wondered why the request was not working.

PR #107, opened 2017-07-29
shutdown: terminate msgStream conns, http conns

### LINKED ISSUE #106: msgStream: error on shutdown
As reported in a private chat by @ronkorving 

```plaintext
<debug> [MAGE processManager] Worker received shutdown request.
<error> [MAGE httpServer] Completed GET /msgstream?transport=longpolling&sessionKey=xxx&rand=0 HTTP/1.1 (500)
  data: {
    "request": {
      "httpVersion": "1.1",
      "method": "GET",
      "url": "/msgstream?transport=longpolling&sessionKey=xxx&rand=0",
      "headers": {
        "host": "localhost:8080",
        "connection": "keep-alive",
        "pragma": "no-cache",
        "cache-control": "no-cache",
        "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.115 Safari/537.36",
        "accept": "*/*",
        "referer": "http://localhost:8080/www/",
        "accept-encoding": "gzip, deflate, br",
        "accept-language": "en-US,en;q=0.8,ja;q=0.6"
      },
      "remoteAddress": "127.0.0.1"
    },
    "response": {
      "status": 500
    },
    "durationSec": 48.175470561
  }
<notice> [MAGE httpServer] HTTP server closed
<debug> [MAGE msgStream] Closing all 1 connection(s)
<emergency> [MAGE processManager] Uncaught exception: Can't set headers after they are sent.
  data: {
    "file": "_http_outgoing.js",
    "line": 371,
    "offset": 11,
    "type": "Error",
    "stack": [
      "ServerResponse.setHeader (_http_outgoing.js:371:11)",
      "ServerResponse.writeHead (_http_server.js:183:21)",
      "HttpPollingHost._sendHeartbeat (node_modules/mage/lib/msgServer/msgStream/transports/http-polling/index.js:220:11)",
      "HttpPollingHost.close (node_modules/mage/lib/msgServer/msgStream/transports/http-polling/index.js:151:7)",
      "MsgStream.close (node_modules/mage/lib/msgServer/msgStream/index.js:119:9)",
      "Server.<anonymous> (node_modules/mage/lib/msgServer/msgStream/index.js:399:8)",
      "emitNone (events.js:91:20)",
      "Server.emit (events.js:188:7)",
      "emitCloseNT (net.js:1577:8)",
      "_combinedTickCallback (internal/process/next_tick.js:77:11)",
      "process._tickDomainCallback (internal/process/next_tick.js:128:9)"
    ]
  }
```

stelcheck:
This commit kills two birds in one stones (apologies to
the families of the birds):

  1. Fixes #106
  2. Tells the msgStream connections to close preemptively;
     this has the advantage of making the shutdown time
     more consistent.

The second point now happens because we immediately emit a 'closing' event on the http server instance; this is required, because the 'close' event emits once all connections are closed, and so listening to that event to close underlying connections is pointless. By closing-and-emitting, we:

  1. Stop other connections from being accepted
  2. Can terminate long-polling and web socket connections when the 'closing' event is emitted
  3. Once we have manually terminated all outstanding long-polling connections, the http server can now close naturally (and fire both the 'close' event and the attached callback)

Tested on the game where the failure was reported (internal).

### DIFF
--- a/lib/msgServer/msgStream/transports/http-polling/index.js
+++ b/lib/msgServer/msgStream/transports/http-polling/index.js
@@ -105,7 +105,7 @@ HttpPollingHost.prototype.deliver = function (msgs) {
 
 
 HttpPollingHost.prototype.respondBadRequest = function (reason) {
-	if (!this.res) {
+	if (this.res.finished) {
 		logger.verbose('Cannot respond bad request (client gone)');
 		return;
 	}
@@ -129,7 +129,7 @@ HttpPollingHost.prototype.respondBadRequest = function (reason) {
 HttpPollingHost.prototype.respondServerError = function () {
 	// used when we fail to confirm a session key for example
 
-	if (!this.res) {
+	if (this.res.finished) {
 		logger.verbose('Cannot respond server error (client gone)');
 		return;
 	}
@@ -191,6 +191,11 @@ HttpPollingHost.prototype.setConnection = function (req, res, query) {
 
 
 HttpPollingHost.prototype.respondToHead = function () {
+	if (this.res.finished) {
+		logger.verbose('Cannot respond to HEAD request (client gone)');
+		return;
+	}
+
 	this.res.writeHead(200, {
 		'content-type': 'application/json',
 		pragma: 'no-cache'
@@ -202,18 +207,17 @@ HttpPollingHost.prototype.respondToHead = function () {
 
 
 HttpPollingHost.prototype._closeConnection = function () {
-	if (!this.res) {
+	if (this.res.finished) {
 		return;
 	}
 
-	this.res = null;
 	this.emit('close');
 	this.removeAllListeners();
 };
 
 
 HttpPollingHost.prototype._sendHeartbeat = function () {
-	if (!this.res) {
+	if (this.res.finished) {
 		return;
 	}
 
--- a/lib/httpServer/transports/http/index.js
+++ b/lib/httpServer/transports/http/index.js
@@ -967,7 +967,13 @@ exports.close = function (cb) {
 
 	logger.debug('Closing HTTP server');
 
+	// Close the server - stop accepting new requests and drain
 	server.close(cb);
+
+	// Let other parts of the system know that the HTTP server
+	// is currently closing; this allows, for instance, for msgStream
+	// to preemptively close current websocket or long-polling connections
+	server.emit('closing');
 };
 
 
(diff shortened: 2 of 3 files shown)


lib/msgServer/msgStream/transports/http-polling/index.js
```diff
@@ -105,7 +105,7 @@ HttpPollingHost.prototype.deliver = function (msgs) {
 
 
 HttpPollingHost.prototype.respondBadRequest = function (reason) {
-	if (!this.res) {
+	if (this.res.finished) {
```
ronkorving: It seems dangerous to me to remove the `(!this.res)` test in these functions.

lib/msgServer/msgStream/transports/http-polling/index.js
```diff
@@ -202,18 +207,17 @@ HttpPollingHost.prototype.respondToHead = function () {
 
 
 HttpPollingHost.prototype._closeConnection = function () {
-	if (!this.res) {
+	if (this.res.finished) {
 		return;
 	}
 
-	this.res = null;
```
ronkorving: I would keep this *and* the finished test. If we try to write twice in the same tick, I'm not sure if `finished` would always be set to the value we would hope it to be.

lib/msgServer/msgStream/transports/http-polling/index.js
stelcheck: I believe the code I replaced it with has the same behavior, except in the cse where we would send before the res is set - which would be something we would definitely want to see pop up as an error I think.

lib/msgServer/msgStream/transports/http-polling/index.js
stelcheck: https://nodejs.org/api/http.html#http_response_finished

```
Boolean value that indicates whether the response has completed. Starts as false. After response.end() executes, the value will be true.
```

ronkorving: Okay, that does sound safe :)

lib/msgServer/msgStream/transports/http-polling/index.js
ronkorving: Yeah, that case doesn't worry me. It was the after-send case that had me worried. But as you point out below, sounds pretty safe. I hope it turns out to be as consistent and predictable as we would need it to be.

PR #112, opened 2017-08-03
state.emit/broadcast: multiple changes

### LINKED ISSUE #91: Data still get sent to the client when there is a server error.
Here is the user commands to reproduce the bug.
```
exports.acl = ['user'];

exports.execute = async function (state, gachaType) {
	const userId = state.actorId;

	state.emit(userId, 'test');

	throw new Error('server');
};
```

stelcheck:
### Behavior changes

  - `state.setTimeout`, if ever used on a user command's state, would leave the client connection hanging forever, if triggered. Since this feature didn't appear to make any sense on `state` in the first place, this feature was re-implemented in the command center instead (see below) 
  - Do not log a warning if we call `emitEvents` without broadcast data
     regardless of whether `msgServer` is present or not
  - Do not log a warning if we call emitEvents without emit data
     regardless of whether the `session` module is present or not
  - Log an error on state close if either of those modules are required, or
    if any other errors are returned

### API changes
  
  - State instanciation now takes an options argument (see IStateOptions)
  - Removed emitToActors (was already deprecated)
  - Removed: state.setTimeout/clearTimeout
  - Removed the 'language' option on emit
  - `emit/broadcast` new `alwaysEmit` option: emit even on failure (fixes #91)
  - New public method `distribute()`: distribute all archivist mutations and events
  - New public method `distributeEvents()`: distribute only events

### Bugfixes

  - `broadcast`: do not send on error (unless alwaysEmit is set) (fixes #91)
  - `emit to self`: do not send on error (unless alwaysEmit is set) (fixes #91)
  - state.emit: do not de-duplicate emitting of the same event to the same actorId multiple times
  
### Internal refactoring
  
  - The state's internal address cache was useless, since archivist was caching values behind that anyway. 
  - StateAccessError -> StateError
  - state.addresse -> state.addressesCache
  - Unit tests: 100% code coverage
  - Updated type definition file

### Other

  - commandCenter: refactoring of the user command execution flow so that all user commands
    would be executed as promises. This helped to simplify the implementation of the `timeout` feature
    (see next point), but also provides a bit more safety, since now early assertions (not executed in an
    asynchronous context) will also be caught 
  - `usercommand.timeout`: configurable user command field allowing to terminate the execution of a
    user command based on a timeout value. Not set by default (so the execution will never time out)

### DIFF
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -79,7 +79,6 @@ function CommandCenter(app) {
 	this.app = app;
 	this.exposed = false;
 	this.commands = {};
-	this.userCommandTimeout = null;
 
 	let ttl;
 
@@ -388,15 +387,52 @@ function compress(content, cb) {
 	});
 }
 
+/**
+ * Promisify the user command execute method, or
+ * ensure its return value is passed to state.respond
+ *
+ * Once callback-based user commands are removed from MAGE,
+ * you will be able to remove this method.
+ *
+ * @param {*} cmdInfo
+ * @param {*} paramList
+ */
+function generateCommandPromise(cmdInfo, state, paramList) {
+	const mod = cmdInfo.mod;
+	const execute = mod.execute.apply.bind(mod.execute, mod, paramList);
+
+	if (cmdInfo.isAsync) {
+		// Async user commands will need to take the response value, and feed it
+		// to `state.respond`
+		return execute()
+			.then((response) => state.respond(response, mod.serialize === false));
+	} else {
+		// Callback-based user commands are expected to call `state.respond`
+		// on their own; all we need to do is promisify
+		return new Promise((resolve) => mod.execute(resolve));
+	}
+}
 
 /**
- * Sets the execution timeout of all user commands to the given duration
+ * Create a promise timer
  *
- * @param {number} timeout   Timeout in milliseconds
+ * Useful when using Promise.race.
+ *
+ * @param {*} timeout
  */
-CommandCenter.prototype.setUserCommandTimeout = function (timeout) {
-	this.userCommandTimeout = timeout;
-};
+function createPromiseTimer(timeout) {
+	let timer;
+
+	const promise = new Promise((resolve, reject) => {
+		if (timeout) {
+			timer = setTimeout(() => reject(new Error('timeout')), timeout);
+		}
+	});
+
+	promise.clear = () => clearTimeout(timer);
+
+	return promise;
+}
 
 /**
  * Returns the information about the given command, or undefined if it does not exist.
@@ -468,24 +504,24 @@ CommandCenter.prototype.buildParamList = function (cmdInfoModParams, cmdParams) 
  * @param {Function} cb
  */
 CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
-	var cmdInfo = this.getCommandInfo(cmd.name);
-	if (!cmdInfo) {
-		// command not registered in this command center
+	const cmdInfo = this.getCommandInfo(cmd.name);
 
+	// Return an error if the user command does not exist
+	if (!cmdInfo) {
 		logger.error('Attempt to execute unregistered user command:', cmd);
 
 		return cb(null, { errorCode: '"server"' });
 	}
 
-	// set up state
-	var state = new State(null, session, {
+	// Set up state
+	const state = new State(null, session, {
 		appName: this.app.name,
 		description: cmd.name,
-		data: metaData,
-		timeout: this.userCommandTimeout
+		data: metaData
 	});
 
-	// check if access control list grants the proper access
+	// Return an error if the user does not have a sufficient right
+	// level to execute this user command
 	if (!state.canAccess(cmdInfo.mod.acl)) {
 		// Create a log entry
 		var logEntry = logger.error.data({
@@ -519,71 +555,62 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 		});
 	}
 
-	var that = this;
-
-	state.archivist.createVault('client', 'client', { state: state }, function () {
-		// execute the command
-
-		logger.debug('Executing user command:', cmdInfo.execPath);
-
-		var mod = cmdInfo.mod;
-
-		var paramList;
+	// Create the client vault, so that events and data may be sent
+	// to the client
+	state.archivist.createVault('client', 'client', { state: state }, () => {
+		// Extract parameters, and return early if the parsing failed
+		let paramList;
 
 		try {
-			paramList = that.buildParamList(cmdInfo.params, cmd.params);
+			paramList = this.buildParamList(cmdInfo.params, cmd.params);
 		} catch (err) {
 			return cb(err);
 		}
 
+		// Debug logging
+		logger.debug
+			.log('Executing user command:', cmdInfo.execPath);
+
+		// Prefix the params list with the state object
 		paramList.unshift(state);
 
-		// time the user command
+		// Set a command timeout, if configured
+		const timer = createPromiseTimer(cmdInfo.mod.timeout);
 
-		var startTime = process.hrtime();
+		// Promisify the user command, if needed
+		const command = generateCommandPromise(cmdInfo, state, paramList);
 
-		// function to run upon completion
-		function onCommandCompleted() {
-			// note: we may not expect an error parameter, the error state should now be known by the state object
+		// Time the user command
+		const startTime = process.hrtime();
 
-			// close the state:
-			// - commits and sends events to other players, or:
-			// - rolls back
+		Promise.race([timer, command])
+			.catch((error) => state.error(error, error))
+			.then(() => {
+				// Clear timeout, if present
+				timer.clear();
 
-			state.close(function (closeError, response) {
-				// use the gathered errors, response, events on the state object to build response JSON for the client
-				// at this time, state.close() never returns a closeError.
+				// Close the state, and respond to the client
+				state.close((closeError, response) => {
+					// use the gathered errors, response, events on the state object
+					// to build response JSON for the client at this time, state.close()
+					// never returns a closeError.
 
-				var durationRaw = process.hrtime(startTime);
-				var duration = durationRaw[0] + durationRaw[1] / 1e9;
+					var durationRaw = process.hrtime(startTime);
+					var duration = durationRaw[0] + durationRaw[1] / 1e9;
 
-				logger.info
-					.data({
-						durationMsec: 1000 * duration,
-						commandName: cmd.name,
-						actorId: state.actorId
-					})
-					.log('Executed user command:', cmd.name);
+					logger.info
+						.data({
+							durationMsec: 1000 * duration,
+							commandName: cmd.name,
+							actorId: state.actorId
+						})
+						.log('Executed user command:', cmd.name);
 
-				exports.emit('completed', that.app, cmd, durationRaw);
+					exports.emit('completed', this.app, cmd, durationRaw);
 
-				cb(null, response);
+					cb(null, response);
+				});
 			});
-		}
-
-		if (cmdInfo.isAsync) {
-			mod.execute.apply(mod, paramList).then(function (response) {
-				state.respond(response, mod.serialize === false);
-			}).catch(function (error) {
-				state.error(error.code, error);
-			}).then(onCommandCompleted);
-		} else {
-			// add the final callback on the params list
-			paramList.push(onCommandCompleted);
-
-			// call the execute function of the usercommand
-			mod.execute.apply(mod, paramList);
-		}
 	});
 };
 
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -478,23 +478,12 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 	}
 
 	// set up state
-
-	var state = new State();
-	state.appName = this.app.name;
-
-	if (this.userCommandTimeout) {
-		state.setTimeout(this.userCommandTimeout);
-	}
-
-	state.setDescription(cmd.name);
-
-	if (session) {
-		state.registerSession(session);
-	}
-
-	if (metaData) {
-		state.data = metaData;
-	}
+	var state = new State(null, session, {
+		appName: this.app.name,
+		description: cmd.name,
+		data: metaData,
+		timeout: this.userCommandTimeout
+	});
 
 	// check if access control list grants the proper access
 	if (!state.canAccess(cmdInfo.mod.acl)) {
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -395,6 +395,7 @@ function compress(content, cb) {
  * you will be able to remove this method.
  *
  * @param {*} cmdInfo
+ * @param {State} state
  * @param {*} paramList
  */
 function generateCommandPromise(cmdInfo, state, paramList) {
@@ -418,9 +419,9 @@ function generateCommandPromise(cmdInfo, state, paramList) {
  *
  * Useful when using Promise.race.
  *
- * @param {*} timeout
+ * @param {number} timeout
  */
-function createPromiseTimer(timeout) {
+function createTimerPromise(timeout) {
 	let timer;
 
 	const promise = new Promise((resolve, reject) => {
@@ -524,7 +525,7 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 	// level to execute this user command
 	if (!state.canAccess(cmdInfo.mod.acl)) {
 		// Create a log entry
-		var logEntry = logger.error.data({
+		const logEntry = logger.error.data({
 			app: this.app.name,
 			module: cmdInfo.mod.name,
 			command: cmd.name,
@@ -568,14 +569,13 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 		}
 
 		// Debug logging
-		logger.debug
-			.log('Executing user command:', cmdInfo.execPath);
+		logger.debug('Executing user command:', cmdInfo.execPath);
 
 		// Prefix the params list with the state object
 		paramList.unshift(state);
 
 		// Set a command timeout, if configured
-		const timer = createPromiseTimer(cmdInfo.mod.timeout);
+		const timer = createTimerPromise(cmdInfo.mod.timeout);
 
 		// Promisify the user command, if needed
 		const command = generateCommandPromise(cmdInfo, state, paramList);
@@ -595,8 +595,8 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 					// to build response JSON for the client at this time, state.close()
 					// never returns a closeError.
 
-					var durationRaw = process.hrtime(startTime);
-					var duration = durationRaw[0] + durationRaw[1] / 1e9;
+					const durationRaw = process.hrtime(startTime);
+					const duration = durationRaw[0] + durationRaw[1] / 1e9;
 
 					logger.info
 						.data({
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -400,17 +400,19 @@ function compress(content, cb) {
  */
 function generateCommandPromise(cmdInfo, state, paramList) {
 	const mod = cmdInfo.mod;
-	const execute = mod.execute.apply.bind(mod.execute, mod, paramList);
+	const execute = () => mod.execute.apply(mod, paramList);
 
 	if (cmdInfo.isAsync) {
 		// Async user commands will need to take the response value, and feed it
 		// to `state.respond`
+		const isJson = mod.serialize === false;
+
 		return execute()
-			.then((response) => state.respond(response, mod.serialize === false));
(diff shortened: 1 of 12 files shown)


christopherobin:
Some of those changes will probably warrant a major release right?

stelcheck:
At the very least a minor , I would assume.

stelcheck:
@christopherobin @ronkorving I would love for this to get merged soon; I am working on https://github.com/mage/mage-module-maintenance which will need this new behaviour to implement maintenance events.

lib/state/state.js
```diff
@@ -302,21 +312,20 @@ State.prototype.emitEvents = function (cb) {
...
-		this.broadcastEvents = undefined;
+	if (events.length > 0) {
 		mage.core.msgServer.broadcast('[' + events.join(',') + ']');
 	}
 
 	events = this.otherEvents;
+	this.otherEvents = undefined;
 
-	if (!events) {
+	if (events === 0) {
```
ronkorving: Isn't `events` an array?

christopherobin: You mean an `array`, I guess `.length` is missing there

mage.ts
```diff
@@ -2415,7 +2415,7 @@ declare namespace mage {
              *
              * @memberOf State
              */
-            emit<T>(actorId: string | string[], eventName: string | number, data: T): void;
+            emit<T>(actorId: string | string[], eventName: string | number, data: T, configuration: mage.core.IStateEmitConfig): void;
```
christopherobin: You might also want to add the `emitEvents` definition since I've seen it used in quite a few projects

stelcheck: I can see cases where this would not behave as intended, since myEvents are currently returned only on close.

What I would suggest is that we create a separate method that will:

  1. Push myEvents into the rest of the event list
  2. Then call emitEvents

@ronkorving would that make sense to you?

stelcheck: Maybe `distributeEvents?`

stelcheck: See the latest commit for the implementation of `distribute` and `distributeEvents`.

lib/state/state.js
```diff
@@ -517,36 +521,72 @@ State.prototype.respond = function (response, isJson) {
...
+ * this state object.
+ *
+ * If you wish to also distribute archivist mutations stacked
+ * on this state, please have a look at the `distribute` method
+ * instead.
+ *
+ * @param {Function} cb callback function
+ */
+State.prototype.distributeEvents = function (cb) {
+	Array.prototype.push.call(this.events, this.myEvents);
```
ronkorving: You can simply:

```js
this.events.push(...this.myEvents);
```

ronkorving: Although that may not be Node 4 compatible.

ronkorving: Just checked, unfortunately it requires Node 6.

ronkorving: Unrelated, but don't you mean to do `apply` instead of `call`? You are currently adding `this.myEvents` as a single element to `this.events`.

stelcheck: Fixed the issues mentioned in your last comment.

ronkorving:
Did this breaking change not cause any documentation changes?

stelcheck:
I have not regenerated the API documentation (since we normally do it on release and these are new APIs), and there are no mentions of `distribute` in the user doc so far.

ronkorving:
What about `state.emit`, which has a breaking change. Those args don't appear anywhere?

stelcheck:
As stated above. We do not use this option in our user doc, and the API doc will be rebuilt on release.

ronkorving:
You were talking about `distribute`, not `emit`, hence my asking :) But we've discussed it offline now, and it's good 👍

stelcheck:
So then the last question becomes, what will now happen of this PR? e.g. should we create minor branches for PRs that are meant for releases in the future?

lib/state/state.js
```diff
@@ -342,9 +351,7 @@ State.prototype.emitEvents = function (cb) {
 			for (var j = 0; j < entries.length; j += 1) {
 				var entry = entries[j];
 
-				if (!entry.language || entry.language === address.language) {
-					actorEvents.push(entry.evt);
-				}
+				actorEvents.push(entry);
```
ronkorving: In other words...

```js
var actorEvents = entries;
```

:D

lib/state/state.js
```diff
@@ -517,36 +511,72 @@ State.prototype.respond = function (response, isJson) {
...
 };
 
+function destroy(error, state, response, cb) {
+	// cleanup
+	state.destroy();
+
+	// return the response
+	if (cb) {
+		cb(null, response);
+	}
```
ronkorving: I'm all for small functions, but this one does the opposite of making code more insightful:

- it does more than just destroy ... its second argument.
- it actually ignores the error argument.
- it only uses `response` and `cb`, just to combine the two (but the caller can't tell).
- it's only ever called from `response`, from where inlining the `destroy` function would have make things a lot clearer I think.

ronkorving: I'm all for small functions, but this one does the opposite of making code more insightful:

- it does more than just destroy ... its second argument.
- it actually ignores the error argument.
- it only uses `response` and `cb`, just to combine the two (but the caller can't tell).
- it's only ever called from `response`, from where inlining the `destroy` function would have make things a lot clearer I think.

lib/state/state.js
```diff
@@ -517,36 +511,72 @@ State.prototype.respond = function (response, isJson) {
...
+}
+
+function respond(error, state, cb) {
+	const myEvents = parseEvents(error, state.myEvents);
+	state.myEvents = undefined;
+
+	destroy(error, state, {
+		response: state.response,
+		myEvents: myEvents.length ? myEvents : undefined
+	}, cb);
```
ronkorving: `cb` ends up being invoked synchronously and never receives any arguments (like error). There's no point in `respond` even receiving `cb` that way.

ronkorving: Also, tracing the `error` argument through these functions, it seems that it's never actually returned in the user command.

lib/state/state.js
stelcheck: Agreed. I'll get that fixed.

lib/state/state.js
stelcheck: No, that error is used to let parseEvents that we wish to retrieve events while we are in an error state.

lib/state/state.js
```diff
@@ -517,36 +481,63 @@ State.prototype.respond = function (response, isJson) {
...
+	state.destroy();
 
-		if (error) {
-			return closeWithError(state, cb);
-		}
+	// return the response
+	if (cb) {
+		cb(null, response);
+	}
+}
```
ronkorving: The `destroy` function serves little (actually.. it serves no) purpose. I would just inline this in `respond`.

lib/state/state.js
```diff
@@ -439,75 +433,45 @@ State.prototype.emit = function (actorIds, path, data, language, isJson) {
...
 
-	if (cb && typeof cb === 'function') {
+	// Log the result
+	logger.error
+		.details(logDetails)
+		.data({
+			actorId: this.actorId,
+			description: this.description
+		})
+		.log('state error');
```
ronkorving: This is quite a breaking change, and not one I'm really happy with. Let's discuss offline.

lib/state/state.js
```diff
@@ -517,36 +482,62 @@ State.prototype.respond = function (response, isJson) {
...
+ * this state object.
+ *
+ * If you wish to also distribute archivist mutations stacked
+ * on this state, please have a look at the `distribute` method
+ * instead.
+ *
+ * @param {Function} cb callback function
+ */
+State.prototype.distributeEvents = function (cb) {
+	this.otherEvents[this.actorId] = this.myEvents.slice();
```
ronkorving: It's possible for `this.actorId` to not be set. You'll probably want to do something there.

But the only 2 things I can currently think of are:

- throw an exception / log an error
- ignore `myEvents`

Neither is very appealing, and we can probably do better by preventing this entire situation from arising by throwing at emit-time. That is, you may emit to `undefined / null / true` if that means emitting to `state.actorId` (as is already the case), but that will throw if `state.actorId` is not set.

Then a check here that either logs or skips `myEvents` if `this.actorId` is not set would become a rarer edge-case (implying the user is really messing about with the system in ways they shouldn't be doing).

stelcheck: `myEvents` can only be empty if there are not `actorId` set, so ignoring would make sense I think.

stelcheck:
Edited the PR's description above https://github.com/mage/mage/pull/112#issue-247677071 - only thing missing at this point is to review the documentation and make the appropriate changes.

stelcheck:
Quickly reviewed. The only thing that will not (yet) be documented is user commands timeout; I have created a separate issue (#134) to deal with that as well as with other missing documentation segments.

lib/commandCenter/commandCenter.js
```diff
@@ -388,15 +387,52 @@ function compress(content, cb) {
...
+ *
+ * Once callback-based user commands are removed from MAGE,
+ * you will be able to remove this method.
+ *
+ * @param {*} cmdInfo
+ * @param {*} paramList
+ */
+function generateCommandPromise(cmdInfo, state, paramList) {
+	const mod = cmdInfo.mod;
+	const execute = mod.execute.apply.bind(mod.execute, mod, paramList);
```
ronkorving: Should this not be `..paramList`? And if so, I'm a bit worried tests didn't catch that.

lib/commandCenter/commandCenter.js
```diff
@@ -388,15 +387,52 @@ function compress(content, cb) {
...
 
+/**
+ * Promisify the user command execute method, or
+ * ensure its return value is passed to state.respond
+ *
+ * Once callback-based user commands are removed from MAGE,
+ * you will be able to remove this method.
+ *
+ * @param {*} cmdInfo
+ * @param {*} paramList
```
ronkorving: `state` is not mentioned in the params

lib/commandCenter/commandCenter.js
```diff
@@ -388,15 +387,52 @@ function compress(content, cb) {
...
+}
 
 /**
- * Sets the execution timeout of all user commands to the given duration
+ * Create a promise timer
+ *
+ * Useful when using Promise.race.
  *
- * @param {number} timeout   Timeout in milliseconds
+ * @param {*} timeout
```
ronkorving: Why remove all typing?

lib/commandCenter/commandCenter.js
```diff
@@ -530,71 +555,62 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
...
 		try {
-			paramList = that.buildParamList(cmdInfo.params, cmd.params);
+			paramList = this.buildParamList(cmdInfo.params, cmd.params);
 		} catch (err) {
 			return cb(err);
 		}
 
+		// Debug logging
+		logger.debug
+			.log('Executing user command:', cmdInfo.execPath);
```
ronkorving: No need for `.log` here, you can call `logger.debug()` directly.

lib/commandCenter/commandCenter.js
```diff
@@ -530,71 +555,62 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
...
+
+		// Prefix the params list with the state object
```
ronkorving: While you're at it, could you pass `state` to `buildParamList` and inject it as the first argument there? `unshift` is very slow, and I don't see why we've never done it in that function instead of unshifting it here.

lib/commandCenter/commandCenter.js
```diff
@@ -530,71 +555,62 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
...
 		paramList.unshift(state);
 
-		// time the user command
+		// Set a command timeout, if configured
+		const timer = createPromiseTimer(cmdInfo.mod.timeout);
```
ronkorving: Name below is `generateCommandPromise`, which I like, but maybe for consistency call this one `createTimeoutPromise` or `createTimerPromise`?

lib/commandCenter/commandCenter.js
```diff
@@ -530,71 +555,62 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
...
+				// Close the state, and respond to the client
+				state.close((closeError, response) => {
+					// use the gathered errors, response, events on the state object
+					// to build response JSON for the client at this time, state.close()
+					// never returns a closeError.
 
-				var durationRaw = process.hrtime(startTime);
-				var duration = durationRaw[0] + durationRaw[1] / 1e9;
+					var durationRaw = process.hrtime(startTime);
+					var duration = durationRaw[0] + durationRaw[1] / 1e9;
```
ronkorving: You've been ES6ifying a lot anyway. You could make these `const`.

lib/commandCenter/commandCenter.js
stelcheck: No, because we cannot use destructuring since we need to support Node 4.

This is why we `bind` to `apply(this, argsArray)`  instead of on `mod.execute` directly.

lib/commandCenter/commandCenter.js
stelcheck: Not sure what types would be valid, but will add State

lib/commandCenter/commandCenter.js
stelcheck: I think this diff compares lines between a function that was removed and a new function but will add the type information.

lib/commandCenter/commandCenter.js
stelcheck: Good catch!

lib/commandCenter/commandCenter.js
stelcheck: Not sure how that would help. `buildParamList` would need to be refactored perhaps?

lib/commandCenter/commandCenter.js
ronkorving: Oh wow. Okay. `apply` *and* `bind`... Alright.

stelcheck: *Yeah...* suggestions welcome though, this looks... bad.

lib/state/state.js
```diff
@@ -7,17 +9,81 @@ var mage;
...
+
+	// Loook up only the addresses which we are unknown to us,
+	// or the ones which are no longer valid
+	var lookup = actorIds.filter((actorId) => {
+		const address = state.addressesCache[actorId];
+
+		return !address || now - address.time >= state.cacheTimeout;
+	});
+
+	mage.session.getActorAddresses(this, lookup, function (error, addresses) {
```
kefniark: this -> state

lib/commandCenter/commandCenter.js
ronkorving: Suggestions...

Either a lambda:

```js
const execute = () => {
  mod.execute(...paramList);
};
```

Or simply not bothering with a shortcut at all:

```js
	if (cmdInfo.isAsync) {
		// Async user commands will need to take the response value, and feed it
		// to `state.respond`
		return mod.execute(...paramList)
			.then((response) => state.respond(response, mod.serialize === false));
	} else {
		// Callback-based user commands are expected to call `state.respond`
		// on their own; all we need to do is promisify
		return new Promise((resolve) => mod.execute(...paramList, resolve));
	}
```

lib/commandCenter/commandCenter.js
```diff
@@ -388,15 +387,53 @@ function compress(content, cb) {
...
+
+	if (cmdInfo.isAsync) {
+		// Async user commands will need to take the response value, and feed it
+		// to `state.respond`
+		return execute()
+			.then((response) => state.respond(response, mod.serialize === false));
+	} else {
+		// Callback-based user commands are expected to call `state.respond`
+		// on their own; all we need to do is promisify
+		return new Promise((resolve) => mod.execute(resolve));
```
ronkorving: Should this not say `execute` instead of `mod.execute`?

lib/commandCenter/commandCenter.js
stelcheck: The first option might be more readable. I'm sold.

lib/commandCenter/commandCenter.js
stelcheck: If `execute` uses `this` then calling just `execute` would break unexpectedly I believe.

stelcheck: Whoops. nope. You are right.

lib/commandCenter/commandCenter.js
```diff
@@ -388,15 +387,55 @@ function compress(content, cb) {
...
  */
-CommandCenter.prototype.setUserCommandTimeout = function (timeout) {
-	this.userCommandTimeout = timeout;
-};
+function createTimerPromise(timeout) {
+	let timer;
+
+	const promise = new Promise((resolve, reject) => {
+		if (timeout) {
+			timer = setTimeout(() => reject(new Error('timeout')), timeout);
```
ronkorving: I think you'll want to add the user command name or state description value to the error message here. Else, when a timeout happens, you won't know on what, right?

lib/state/state.js
```diff
@@ -7,17 +9,44 @@ var mage;
...
 var NO_DESCRIPTION = 'no description';
 
-function StateAccessError(message, details) {
+/**
+ * State error class
+ *
+ * @param {*} message
+ * @param {*} details
+ */
+function StateError(message, details) {
```
ronkorving: Not a blocker at all, especially since this already existed. But ES6 classes work miracles when extending `Error`.

lib/state/state.js
```diff
@@ -7,17 +9,44 @@ var mage;
...
+/**
+ * Default description if none is set on a state object
+ */
 var NO_DESCRIPTION = 'no description';
 
-function StateAccessError(message, details) {
+/**
+ * State error class
+ *
+ * @param {*} message
```
ronkorving: is a string.

lib/state/state.js
```diff
@@ -7,17 +9,44 @@ var mage;
...
+ * @param {*} details
```
ronkorving: is an object

lib/state/state.js
```diff
@@ -26,75 +55,85 @@ util.inherits(StateAccessError, Error);
...
 	this.myEvents = [];           // array of JSON serialized events for this session
 
-	this.addresses = {};            // { actorId: { actorId, language, addrName, clusterId }, actorId: false }
-	this.otherEvents = undefined;	// events for other actors: { actorId: [ list of events ], actorId: etc... }
-	this.broadcastEvents = undefined;
-
-	this.timeout = null;	// timeout
-	exports.emit('created');
-}
+	this.addressesCache = {};            // { actorId: { actorId, language, addrName, clusterId }, actorId: false }
```
ronkorving: `addressesCache` does not appear elsewhere.

lib/commandCenter/commandCenter.js
stelcheck: `state.error` logs description - which is set to the name of the user command.

lib/state/state.js
```diff
@@ -292,76 +305,76 @@ State.prototype.findActors = function (actorIds, cb) {
...
+function parseEvents(isError, events) {
+	return events
+		.filter((entry) => !isError || entry.alwaysEmit)
+		.map((entry) => entry.evt);
+}
 
-	// this function emits all events to players who are not state.actorId
 
-	events = this.broadcastEvents;
+State.prototype.emitEvents = function (isErrorState, cb) {
```
ronkorving: Not a blocker, but might be nice to underscore this.

lib/state/state.js
```diff
@@ -292,76 +305,76 @@ State.prototype.findActors = function (actorIds, cb) {
...
-	this.lookupAddresses(actorIds, function (error, addresses) {
+	lookupAddresses(this, actorIds, function (error, addresses) {
 		if (error) {
 			return cb(error);
 		}
 
 		for (var i = 0; i < actorIds.length; i += 1) {
 			var actorId = actorIds[i];
-			var entries = events[actorId];
+			var entries = parseEvents(isErrorState, events[actorId]);
```
ronkorving: Since `parseEvents` has a relevant cost, it might be good to `continue` early if there is no `address` for this `actorId`, which will be *very* common. We very often emit events to a user's "friends" etc, without knowing if that friend is online. Assuming most of the day users are *not* playing, we will have more misses than hits.

lib/state/state.js
```diff
@@ -375,14 +388,13 @@ function serializeEvent(path, data, isJson) {
 }
 
 
-State.prototype.broadcast = function (path, data, isJson) {
-	var evt = serializeEvent(path, data, isJson);
+State.prototype.broadcast = function (path, data, options) {
```
ronkorving: I guess this is worthy of JSDoc :)

lib/state/state.js
```diff
@@ -399,115 +411,75 @@ function isSelf(state, actorId) {
 }
 
 
-State.prototype.emit = function (actorIds, path, data, language, isJson) {
-	// language may be omitted
-	// if language is provided, it will only emit if the language matches the language used by actorId
-
-	var len;
+State.prototype.emit = function (actorIds, path, data, options) {
```
ronkorving: I guess this is worthy of JSDoc :)

lib/state/state.js
stelcheck: TBH I now tend to put user doc only in the typedoc file. We might want to discuss this soon.

lib/state/state.js
stelcheck: Fixed

PR #114, opened 2017-08-04
details on vault setup errors, create migrates

### LINKED ISSUE #104: Cryptic error if you haven't run npm run archivist-create
The error seems to come from auth in the mage server console window, and invalidusername on client side.

stelcheck:
Error log will hint the user that they need to run `archivist:create`
to setup up the storage.

`archivist:create` now implies `archivist:migrate`.

Fixes #104 cc @tatsujinichi

### DIFF
--- a/lib/archivist/configuration.js
+++ b/lib/archivist/configuration.js
@@ -132,6 +132,14 @@ function createVault(vaultName, vaultType, vaultConfig, cb) {
 
 		vault.setup(vaultConfig || {}, function (error) {
 			if (error) {
+				logger.alert
+					.details('Early errors may mean one of the following:')
+					.details('  1. The remote server is down or inaccessible')
+					.details('  2. You need to run `npm run archivist:create` to create')
+					.details('     and configure storage for your vault backend')
+					.details('See error data for more details.')
+					.log(`Failed to set up vault ${vaultName}`);
+
 				return cb(error);
 			}
 
--- a/lib/tasks/archivist-create.js
+++ b/lib/tasks/archivist-create.js
@@ -1,14 +1,11 @@
 var async = require('async');
+var migrate = require('./migrate');
 
 
 exports.setup = function (mage, options, cb) {
 	async.series([
-		function (callback) {
-			mage.core.loggingService.setup(callback);
-		},
-		function (callback) {
-			mage.core.archivist.setup(callback);
-		}
+		(callback) => mage.core.loggingService.setup(callback),
+		(callback) => mage.core.archivist.setup(callback)
 	], cb);
 };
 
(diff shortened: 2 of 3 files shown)


stelcheck:
cc @ronkorving

stelcheck:
Just noticed that this won't work at least for couchbase, since we do not connect until we make the first request; I would suggest changing this behavior and connecting as part of the setup instead, but then MAGE would crash at startup instead of retrying (I think?).

stelcheck:
A quick investigation of how vault backends behave:

  - Couchbase: does nothing (assuming it retries to connect?)
  - Memcached: idem
  - Redis: tries to connect forever, and emits an emergency error to log

So at this point, I am assuming that vault backends not connecting or databases/buckets not being found would just emit errors to log in separate place.

Suggestions on how to handle this correctly more than welcome.

stelcheck:
Ping @mage/mage-core-team - any ideas how we could crack this one?

ronkorving:
Here's how I would personally want all vaults to behave:

**Setup responsibility**

The asynchronous setup phase connects to and confirms that the database is accessible and ready to be used.

**Error handling**

- If it cannot connect because there's nothing listening on the port, enter a retry-loop for up to N seconds. Ideally, this also applies to TCP connections being cut off (during containerization).
- If it cannot connect because the database server returns an *expected* error that indicates waiting will result in success (like Couchbase not being ready for N1QL queries yet), enter a retry-loop for up to N seconds.
- If it cannot connect because the database server returns an *unexpected* error, fail immediately with an emergency-level log.
- If a retry-loop times out, fail with an emergency-level log message.

If we can agree on these rules, we should document them and go through every vault to make sure this is how they behave or fix them accordingly.

stelcheck:
Sounds reasonable to me, but that means *every* vault backends would need some code to deal with that... may take some time to get that set up. Also, from the setup phase, do we have a way to know, for instance, which bucket/database are supposed to be present, and a mean to confirm whether migration was run and is up-to-date?

ronkorving:
1. Which buckets/databases are supposed to be present: yes
2. How to confirm whether migration is up-to-date: yes

Re. 1: each vault connects to one bucket/database and it knows which to connect to.
Re. 2: every bucket/database stores migration state, so it can be queried.

stelcheck:
Thanks for the clarifications.

How would we know the expected database migration version number by the code though? I do not believe there is an API for that yet (although it may be trivial to add).

stelcheck:
Moving this to the 1.3.0 milestone, since it will likely take quite some time to safely complete should we follow this approach.

Nokitoo:
mage/mage#234 already solved this issue

PR #115, opened 2017-08-04
Dynamic configuration

### LINKED ISSUE #67: config: enforce configuration values programmatically
For testing purposes, I would like to make sure to programmatically disable certain modules such as serviceDiscovery and MMRP - instead of having to enforce the use of an environment variable for running tests. 

@ronkorving @qubyte help/pointers on how I could get this done would be most welcome.

### LINKED ISSUE #124: Custom config archivist entries don't seem to override default config.
npm install mage --create --typescript --prefix my-gameserver

> config/custom.yaml

```yaml
archivist:
    vaults:
        # Our vaults, augment this with as many as you need.
        # You should take care to name them appropriately given the context of how they will be used.
        volatileVault:
            type: memory
        userVault:
            type: file
            config:
                path: ./filevault/userVault
        tableVault:
            type: file
            config:
                path: ./filevault/tableVault

    # When doing "list" operations, we will attempt each mentioned vault until successful
    listOrder:
        - userVault
        - tableVault

    # When doing "get" operations, we will attempt each mentioned vault until successful
    readOrder:
        - volatileVault
        - userVault
        - tableVault

    # When doing "add/set/touch/del" operations, we will write to each mentioned vault in the given order
    writeOrder:
        - client
        - volatileVault
        - userVault
        - tableVault
```

add to file

>lib/archivist/index.ts
```javascript
exports.table = <mage.archivist.ITopic> {
  index: ['tableId'],
  vaults: {
    tableVault: {
    }
  }
};
```

Then run `npm run archivist-create`;

```
m-18602 - 16:26:19.071     <debug> [MAGE archivist] Vault "tableVault" is not in list/read/write-order, skipping.
m-18602 - 16:26:19.075 <emergency> [MAGE archivist] None of the mentioned vaults for topic "table" is available.
  data: {
    "file": "configuration.js",
    "line": 620,
    "offset": 10,
    "type": "Error",
    "stack": [
      "node_modules/mage/lib/archivist/configuration.js:620:10",
      "Array.forEach (native)",
      "registerTopics (node_modules/mage/lib/archivist/configuration.js:596:9)",
      "Object.exports.setup (node_modules/mage/lib/archivist/configuration.js:675:3)",
      "EventEmitter.exports.setup (node_modules/mage/lib/archivist/index.js:20:16)",
      "setupArchivist (node_modules/mage/lib/tasks/serve.js:23:23)",
      "node_modules/async/lib/async.js:610:21",
      "node_modules/async/lib/async.js:249:17",
      "iterate (node_modules/async/lib/async.js:149:13)",
      "node_modules/async/lib/async.js:160:25"
    ]
  }
```

Copying the config from the custom.yaml to default.yaml fixes this.

I expect config in the custom config to override or add to the default config.

stelcheck:
Enable developers to dynamically change the configuration at runtime.
Includes documentation and a test.

### Todo

 - [x] Complete test suite (make sure dynamic config overrides defaults, file and environment)
 - [x] Type definition file update

Fixes #67
Fixes #124


lib/mage/Mage.js
```diff
@@ -222,6 +222,10 @@ Mage.prototype.getTask = function () {
 Mage.prototype.setupCoreLibs = function () {
 	var core = this.core;
 
+	// Load the configuration
+
+	core.config = require('../config');
```
ronkorving: If this throws or wants to log for any reason, will it be able to?

lib/config/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+ * See: https://mage.github.io/mage/#dynamic-configuration
+ */
+export = {
+    /**
+     * Programatically set a configuration entry
+     *
+     * @param key dot-separated configuration path (example: 'server.hosts')
+     * @param value
+     */
+    set(key: string, value: any): void;
```
ronkorving: No `get`?

lib/mage/Mage.js
stelcheck: You mean the require?

lib/config/index.d.ts
stelcheck: get should be accessed through mage.core. Not sure we need to expose it here.

lib/mage/Mage.js
ronkorving: Yes

stelcheck: Not sure what you mean by "this" - if you are asking whether config will be able to log, then the answer is yes.

stelcheck:
This should now be ready to go.

lib/mage/Mage.js
ronkorving: If the require call throws (because at the top-scope of `../config` something goes wrong), will that be logged? There is no logger on `mage.core` yet.

stelcheck: https://github.com/mage/mage/pull/115/files#diff-5122ec46a4f7ff0f1909a379117ff89bR6 and some code a bit below should provide config with a default logger. So I *think* this is enough, but let me know if it isn't.

ronkorving: LGTM 👍

stelcheck:
Added proper support for custom config to this PR.

docs-sources/includes/_Configuration.md
```diff
@@ -55,6 +66,26 @@ To run your game in development, MAGE has a `developmentMode` configuration flag
...
+```yaml
+server:
+    clientHost:
+      bind: MAGE_APP_BIND:string
+
+    mmrp: MAGE_APP_MMRP:bool
+```
+
+You may also add a `config/environment.yaml` file to your project: this file serves
+as a mean to connect environment variables to specific configuration entries. Environment
```
ronkorving: "a means" I think
Maybe a native English speaker can confirm

stelcheck: I ran this into Grammarly, and it did not return any errors or suggested any changes.

lib/config/config.js
```diff
@@ -12,10 +13,21 @@ var logger;
...
 var configList = [];
 
+var runtimeConfig = new Matryoshka({}, 'runtime');
+
 var aggregate;
 
 var isInitialized = false;
 
+// Set up the logger with a nice default before we have access to config;
+// This will be overriden by MAGE once during its instanciation, but
```
ronkorving: I see a lot of people do this, but ..
https://en.wiktionary.org/wiki/instanciation

lib/config/config.js
```diff
@@ -234,6 +246,35 @@ exports.setDefaults = function (sourcePath) {
 
 
 /**
+ * Dynamically set a configuration value
+ *
+ * @param {String} name       The key to assign the content to.
+ * @param {*} value 		  Value
```
ronkorving: tab whitespace messed up alignment

lib/config/config.js
```diff
@@ -234,6 +246,35 @@ exports.setDefaults = function (sourcePath) {
 
 
 /**
+ * Dynamically set a configuration value
+ *
+ * @param {String} name       The key to assign the content to.
```
ronkorving: suggest: content -> value

lib/config/config.js
```diff
@@ -234,6 +246,35 @@ exports.setDefaults = function (sourcePath) {
...
+ * @param {*} value 		  Value
+ */
+exports.set = function (name, value) {
+	const obj = {};
+	const subpaths = name.split('.');
```
ronkorving: In `exports.get`, the name (ie: `trail`) is allowed to be an array I think. Might be nice to support that in `set` too.
Also, `trail` would be a more consistent name than `name` given the rest of this module.

lib/config/config.js
```diff
@@ -309,13 +350,17 @@ exports.getMatryoshka = function (trail) {
...
 	if (isInitialized) {
 		throw new Error('The config module has already been initialized.');
 	}
 
 	isInitialized = true;
-	logger = logObj;
+
+	// We create a temporary log creator until MAGE provides us with one;
+	// See `setLogger` for more details.
+	const temporaryLogger = loggingService.createLogCreator();
```
ronkorving: The point of dependency injection through `initialize` was to make the module more testable. I don't know what would be an elegant alternative to what you've done here though. In any case, injecting the logger through `setLogger` is more elegant as far as I'm concerned than passing it to an `initialize` function, but that's kinda besides the point.

So, not a blocker, but maybe you have some ideas?

stelcheck: I don't really see a current purpose to making this injectable... if we had to, I guess we could externalize that part into a function.

ronkorving: We can do that if/when it becomes an issue 👍

PR #117, opened 2017-08-16
Update RELEASE_GUIDE.md

ronkorving:
I've added a how-to for cherry-picking a patch-release. I've not fully tested it yet, to avoid noise on the GitHub repo. Let's start by picking this apart :)

I can imagine there may be some superior (and/or simpler) git commands that can be used to visualize changes between branches that someone may be aware of. I would love to hear it if that's the case :)

### DIFF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -45,7 +45,7 @@ To make a patch-release, we cherry-pick commits from master into a temporary bra
 5. Run: `git push upstream 1.2.4` (replace `1.2.4` by the new version tag)
 6. Write up the release notes at the new tag on https://github.com/mage/mage/releases/new
 7. Run: `npm publish`
-8. Remove your temporary release branch: `git checkout master && git branch -d release`
+8. Remove your temporary release branch: `git checkout master; git branch -d release`
 
 ## After publishing the release
 
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -10,15 +10,42 @@ We apply a semver-ish versioning scheme, using the following logic:
 
 ## How to release
 
-1. Ensure your shell is in the `master` branch
-2. Ensure you are up-to-date: `git pull upstream master` (where "upstream" references https://github.com/mage/mage)
-3. Determine whether this release is a major, minor or patch release
-4. Run: `npm version major`, `npm version minor` or `npm version patch`. This will have:
+In all the `git` shell commands below, `upstream` references the repository at https://github.com/mage/mage.
+
+1. Ensure your shell is in the `master` branch: `git checkout master`
+2. Ensure you are up-to-date: `git pull upstream master`
+3. Continue below
+
+### A) Minor or Patch release based on master
+
+If this is a patch release or if you are making a minor release that is equal to the current state of the `master` branch, you can release the `master`-branch directly.
+
+1. Compare changes with the previous release: `git log master...1.2.3` (replace `1.2.3` by the previous version tag)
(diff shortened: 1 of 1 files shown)


RELEASE_GUIDE.md
```diff
@@ -10,15 +10,42 @@ We apply a semver-ish versioning scheme, using the following logic:
...
-2. Ensure you are up-to-date: `git pull upstream master` (where "upstream" references https://github.com/mage/mage)
-3. Determine whether this release is a major, minor or patch release
-4. Run: `npm version major`, `npm version minor` or `npm version patch`. This will have:
+In all the `git` shell commands below, `upstream` references the repository at https://github.com/mage/mage.
+
+1. Ensure your shell is in the `master` branch: `git checkout master`
+2. Ensure you are up-to-date: `git pull upstream master`
+3. Continue below
+
+### A) Minor or Patch release based on master
```
stelcheck: I would suggest clearly separating major/minor releases (where we release everything) from patch releases.

Patch releases may happen while we have already merged content meant for the next minor/major, so we would need to cherry-pick in the same way as if we had to cherry-pick for a backport. Sometimes you will want to merge all fixes because you have no undesired commits, but it's probably easier not to think about this edge case for documentation purposes.

RELEASE_GUIDE.md
```diff
@@ -10,15 +10,42 @@ We apply a semver-ish versioning scheme, using the following logic:
...
-5. Run: `git push upstream master --tags` (where "upstream" references https://github.com/mage/mage)
+3. Run: `git push upstream master --tags`
+4. Write up the release notes at the new tag on https://github.com/mage/mage/releases/new
+5. Run: `npm publish`
+
+### B) Patch release based on a previous release
+
+If we have already merged minor-level commits into `master` that should not land in this patch release, we can cherry-pick our way to a patch release.
+
+1. Create a release branch: `git checkout -b release 1.2.3` (replace `1.2.3` by the previous version tag)
```
stelcheck: This will fail if I have done a prior release. Would recommend:

```shell
git checkout 1.2.3
git checkout -b releases/1.2.3
```

We could do it in a single command, of course, I just find two lines more readable (personal preference).

RELEASE_GUIDE.md
```diff
@@ -10,15 +10,42 @@ We apply a semver-ish versioning scheme, using the following logic:
...
+2. Compare changes with `master`: `git log master...release --cherry-pick` (this is where we will pick commits from)
+3. Pick all commits you wish to include: `git cherry-pick COMMITHASH` (replace `COMMITHASH`)
+4. Run: `npm version patch`. This will have:
+    1. Updated the version in package.json
+    2. Regenerated documentation
+    3. Committed (git)
+    4. Tagged (git)
+5. Run: `git push upstream --tags`
```
stelcheck: Depending on git configuration, this may do nothing or break. See https://stackoverflow.com/a/948397/262831

RELEASE_GUIDE.md
ronkorving: That command worked for me?
It makes a branch based called `release` based on what `1.2.3` refers to.

stelcheck: Like I said. Personal preference. Doing it on one line is perfectly fine.

ronkorving: I understand preference, but you mentioned this can fail. I didn't quite understand under what circumstances and why.

stelcheck: The failure I mentioned was about attempting to re-create the same branchore than once, and was unrelated to splitting the command on two.

PR #125, opened 2017-08-22
Remove auth module, document how to enable it

### LINKED ISSUE #58: Bootstrap creates filevaults for auth not in gitignore
Noticed this when about to push a small sample project, the gitignore doesn't include the file vaults.  After a conversation with @ronkorving it might not be appropriate to create the auth module and such in the bootstrap.

I'm on the side that creating these is actually a good thing, but it might take the user down the path where they use filevault for everything, like auth and such, which probably isn't so good.

@stelcheck what is your opinion on this? How do you think we should go about it.

christopherobin:
Fixes #58

### DIFF
--- a/docs-sources/includes/_Actors.md
+++ b/docs-sources/includes/_Actors.md
@@ -9,20 +9,62 @@ is assigned a unique session ID.
 As long as a session ID is used and reused by an actor, it will stay active. After a long period
 of non-activity however, the session will expire and the actor will be "logged out" as it were.
 
-## Auth and Session modules
+## Session module
 
 > `lib/index.js`
 
 ```javascript
 mage.useModules([
-  'auth',
   'session'
 ]);
 ```
 
-Freshly bootstrapped MAGE applications already have the auth and the session module activated and
-configured (including a basic Archivist configuration which will store auth-information to disk
-and session-information in memory).
+Freshly bootstrapped MAGE applications already have the session module activated and configured
+(including a basic Archivist configuration which will store session-information in memory).
+
+## Auth module
+
+> `lib/index.js`
+
+```javascript
+mage.useModules([
+  'auth'
+]);
+```
+
+The `auth` module is not enabled by default since it requires a working archivist topic
+to store user data. To enable it make sure you have working vaults then add the following
+to your application:
+
+> lib/archivist/index.js
+
+```javascript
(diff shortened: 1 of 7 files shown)


docs-sources/includes/_Actors.md
```diff
@@ -9,20 +9,62 @@ is assigned a unique session ID.
...
-  'auth',
   'session'
 ]);
 ```
 
-Freshly bootstrapped MAGE applications already have the auth and the session module activated and
-configured (including a basic Archivist configuration which will store auth-information to disk
-and session-information in memory).
+Freshly bootstrapped MAGE applications already have the session module activated and configured
+(including a basic Archivist configuration which will store session-information in memory).
```
stelcheck: How much work would it be to add a paragraph on how to wire this into another vault backend?

docs-sources/includes/_Actors.md
```diff
@@ -9,20 +9,62 @@ is assigned a unique session ID.
...
+
+> `lib/index.js`
+
+```javascript
+mage.useModules([
+  'auth'
+]);
+```
+
+The `auth` module is not enabled by default since it requires a working archivist topic
```
stelcheck: Just saying 'To enable this module...' should be sufficient, I think.

docs-sources/includes/_Actors.md
```diff
@@ -9,20 +9,62 @@ is assigned a unique session ID.
...
+to store user data. To enable it make sure you have working vaults then add the following
+to your application:
```
stelcheck: Keep in mind, these code blocks are appearing on the right, not below. You should make sure to generate the doc to see the final render of the content.

docs-sources/includes/_Actors.md
```diff
@@ -31,6 +73,9 @@ and session-information in memory).
 ```javascript
 exports.login = function (state, username, password, callback) {
   mage.auth.login(state, username, password, callback);
+
+  // if you are not use the auth module, auth can be done manually by calling
```
stelcheck: It would be wiser to just tell people to use auth I think, or to actually provide a full-blown example on how to do this manually.

docs-sources/includes/_Actors.md
christopherobin: Added a small paragraph below, also explaining why

PR #126, opened 2017-08-22
Updated documentation

christopherobin:
* Moved table creation explanation from SQLite3 to MySQL and linked them
* Added line breaks in several part of the documentation to prevent the code samples from bunching up

### DIFF
--- a/docs-sources/includes/_Archivist.md
+++ b/docs-sources/includes/_Archivist.md
@@ -130,6 +130,13 @@ config:
 
         # optional, useful if you share a bucket with other applications
         prefix: "bob/"
+
+    # options only used with archivist:create
+    create:
+        adminUsername: admin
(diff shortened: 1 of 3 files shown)


docs-sources/includes/_Archivist.md
```diff
@@ -130,6 +130,13 @@ config:
 
         # optional, useful if you share a bucket with other applications
         prefix: "bob/"
+
+    # options only used with archivist:create
+    create:
```
stelcheck: The indent looks off.

christopherobin: No, `create` is outside of `options`

christopherobin: Not sure exactly why, I guess because options is partially feed directly to some of the couchbase functions?

PR #127, opened 2017-08-22
npm command format

### LINKED ISSUE #118: mage command should use the same format as npm commands
For `npm run` we have migrated toward a `category:command` format, like `archivist:create`. But for the `mage` binary we are still using the `archivist-create` format which makes documenting inconsistent (parts of the documentation uses the new syntax and others the old syntax).

It seems like the recommended solution would be to convert the `mage` sub-commands to use the same format as npm.

stelcheck:
- use : instead of - as a context separator
  - update the documentation wherever needed

Fixes #118


lib/cli/index.js
```diff
@@ -15,37 +15,37 @@ program.option('--stack-limit <limit>', 'change the stack trace limit', parseInt
 // tasks
 
 program
-	.command('show-client-config')
+	.command('show:client-config')
```
christopherobin: How about changing this to `show:config:client` ?

christopherobin: Also it feels like this command requires an app, why not make it an argument like `show:config`?

stelcheck: cc @ronkorving

ronkorving: @christopherobin It does require an app, see the `--app` argument 2 lines below.

stelcheck: What we should try to aim at here is semantic separation by categories, I think - meaning that `-` should be a word separator and `:` should be a category separator.

With that in mind, I believe we have the following options:

  1. Leave it as is, and consider that `show should normally be a flat category (no subcategory)
  2. Classify show by sub-category (`show:server:*`, `show:client:*`). 
  3. Put all client commands under a completely separate root category `client` (`client:show:config`, `show:config`)

The question is, which one is most natural to end-users, and which one will keep making sense as we add/remove categories in the future.

christopherobin: @ronkorving The thing is that AFAIK `--app` is required, if you don't pass it the command will fail. So for me `app` should instead be an argument since flags are supposed to be used for modifying execution and not providing critical data imo.

ronkorving: I would add that categorization may actually suggest another option: `config:show`. Thoughts?

christopherobin: How about `config:server` and `config:client` then?

stelcheck: Sounds reasonable. Let's go with that.

stelcheck:
PR updated. Removed the Readme from `lib/cli` since it didn't make much sense any more.

stelcheck:
@christopherobin @ronkorving can either of you give the final approval, please?

christopherobin:
I still would prefer if `--app` would be an argument since like I said not passing any `--app` will cause the command to fail, otherwise looks goods

stelcheck:
@ronkorving @christopherobin fixed!

ronkorving:
@stelcheck I may be missing something, but I don't see a change re. `--app`.

stelcheck:
Have a look at each template’s package.json

ronkorving:
The way I interpreted @christopherobin 's feedback was: turn `npm foo --app bar` into: `npm foo bar` since it's not an option? I'll shut up now and let @christopherobin speak for himself though :)

christopherobin:
This is exactly what I mean, to make it more concrete, the current behavior would look like this if implemented on ls:

```bash
$ ls roll/
error: a folder is required
$ ls --folder roll/
. .. never gonna give you up
```

IMO It just feels wrong, now feel free to do what you want.

stelcheck:
@christopherobin @ronkorving I am confused. Do we want:

```
npm run config:client game
```

or

```
npm run config:client
```

christopherobin:
`npm run config:client game`, with the usage documentation showing it as:

```bash
$ mage config:client
emergency: missing app name
$ mage ---help

  Usage: mage [options] [command]

  Commands:

    config:client [app_name]         output the full client configuration in JSON for the given app name
    config:server [options] [trail]  output the full configuration, or the sub-config at the given trail in JSON

... snip
$ mage config:client game
{"some":["json","output"],"for":"game"}
```

stelcheck:
@christopherobin ok, the behavior should be OK now. One last thing: I don't *think* we can currently walk around *not* using the mage logger, unless we `process.exit` without calling the callback - which I am not sure we should do at this point, this wasn't the point of this PR anyway (but would likely still be something we should look into).

christopherobin:
The mage logger is fine in this case, I just didn't feel like copying the whole logger output in my sample.

stelcheck:
Great! Could anyone approve in this case, so we can merge?

stelcheck:
... sorry.

PR #128, opened 2017-08-22
Right-pad PID in terminal and file logger

### LINKED ISSUE #123: Terminal logger: PID causes indent discrepancies
```text
w-3193 - 21:35:05.128    <notice> [MAGE httpServer] Server running at http://0.0.0.0:8080
w-3193 - 21:35:05.129     <debug> [MAGE processManager] Startup phase completed in 2.243 sec
m-12551 - 21:35:05.133     <debug> [MAGE mmrp] ionring:12551 received handshake from ["ionring:12551:3193:1"]
w-3193 - 21:35:05.664     <debug> [MAGE REPL] connecting to master process
m-12551 - 21:35:05.665    <notice> [MAGE REPL] connected
```

While both Windows and Linux do not have theoretical limits I could find, defaults seem to indicate a standard max PID of between 32768 and 65535. Therefore, I think it would be safe to right-pad PIDs less than 5 numbers long, just to get everything even.

stelcheck:
Fixes #123

### DIFF
--- a/lib/loggingService/writers/file.js
+++ b/lib/loggingService/writers/file.js
@@ -171,7 +171,14 @@ FileWriter.prototype.channelFunctionGenerator = function (channelName) {
 
 	var jsonIndent = this.jsonIndent;
 	var role = isMaster ? 'm' : 'w';
-	var prefix = role + '-' + process.pid + ' - ';
+	var prefix = '';
+
+	// Attach the role to the prefix
+	prefix += role + '-';
+
+	// Attach the padded PID to the prefix
+	// See: https://github.com/mage/mage/issues/123
+	prefix += ('' + process.pid + '     ').slice(5) + ' - ';
 
 	// channel hint reset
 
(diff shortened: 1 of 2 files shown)


lib/loggingService/writers/file.js
```diff
@@ -171,7 +171,14 @@ FileWriter.prototype.channelFunctionGenerator = function (channelName) {
...
 	var role = isMaster ? 'm' : 'w';
-	var prefix = role + '-' + process.pid + ' - ';
+	var prefix = '';
+
+	// Attach the role to the prefix
+	prefix += role + '-';
+
+	// Attach the padded PID to the prefix
+	// See: https://github.com/mage/mage/issues/123
+	prefix += ('' + process.pid + '     ').slice(5) + ' - ';
```
ronkorving: Why the prefix empty string?
(a bit late, sorry)

stelcheck: To make sure we are always concatenating a string.

ronkorving: The postfix string ensures it.

PR #132, opened 2017-08-28
migrations: do not use package.json ver. as target

### LINKED ISSUE #121: migration: `package.json` version should not be relevant
Ref: https://github.com/mage/mage/pull/119#discussion_r133621308

I do not believe migration versioning has ever been used in such a way. Further more; I think it may be easier to version schemas separately from the application's version.

stelcheck:
Fixes #121

### DIFF
--- a/docs-sources/includes/_Archivist.md
+++ b/docs-sources/includes/_Archivist.md
@@ -598,11 +598,10 @@ It is not recommended to use those features in production.
 ## Migrations
 
 MAGE supports database migration scripts similar to [Ruby on Rails 2.1](http://api.rubyonrails.org/classes/ActiveRecord/Migration.html)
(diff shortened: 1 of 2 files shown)


stelcheck:
@ronkorving @christopherobin this may need additional discussions before merging. Please let me know your thoughts on this PR.

stelcheck:
One issue I can see is that now, we could version each schema entirely separately, so I am not sure if passing a version number to `archivist:migrate` would ever make sense.

ronkorving:
@stelcheck True, but if you extend that, migrating down will also never make sense anymore. Down to what? The entire concept was there to allow rollback after a botched deployment. In my memory, we've never had to resort to that, but that was the purpose of all this.

stelcheck:
Hmm, you are right. You would need to roll back before reverting your code base. So yes, this is still required.

PR #138, opened 2017-09-01
daemon: multiple fixes

### LINKED ISSUE #136: `npm reload` crashes
```
Reloading... (pid: 8771)
m-8844  - 14:59:24.103 <emergency> [MAGE processManager] Uncaught exception: Cannot read property 'pipe' of undefined
  error: {
    "type": "TypeError",
    "file": "index.js",
    "line": 147,
    "offset": 13,
    "stack": [
      "process.sigFailedHandler (node_modules/mage/lib/daemon/index.js:147:13)",
      "emitNone (events.js:86:13)",
      "process.emit (events.js:188:7)",
      "Signal.wrap.onsignal (internal/process.js:215:44)"
    ]
  }
```

### LINKED ISSUE #137: `npm run reload` causes heapdumps
Both the `heapdump` module and `reload` share the same signal, so triggering one always triggers the others. This means that one cannot take heapdumps without triggering reloads.

stelcheck:
Demo video: https://asciinema.org/a/ASEQCfwIdNzFZErCGhOTXi9w6

  - `reload` SIGUSR2 -> SIGALARM (fixes #137)
  - `reload`: reverting to `spawn` instead of of `fork` (fixes #136)
  - Using a named pipe for sharing logs between MAGE and the daemonized
  - start, stop, restart and reload show logs
  - all logs are always shown (instead of only errors on crashes)
  - colorization is forced when daemonized

### DIFF
--- a/lib/daemon/index.js
+++ b/lib/daemon/index.js
@@ -1,10 +1,15 @@
 // Command manager for daemonization
 
-var fs = require('fs');
-var path = require('path');
-var cluster = require('cluster');
-var interrupt = require('./interrupt');
 var chalk = require('chalk');
+var cluster = require('cluster');
+var fs = require('fs');
+var net = require('net');
+var readline = require('readline');
+var path = require('path');
+var rimraf = require('rimraf');
+var xpipe = require('xpipe');
+
+var interrupt = require('./interrupt');
 var mage = require('lib/mage');
 
 
@@ -12,11 +17,41 @@ var mage = require('lib/mage');
 
 mage.core.config.setTopLevelDefault('daemon', path.join(__dirname, 'config.yaml'));
 
-var APP_LOCK_FILE = mage.core.config.get(['daemon', 'pidfile']);
-var CMD_LOCK_FILE = './.commandpidfile';
+const APP_LOCK_FILE = mage.core.config.get(['daemon', 'pidfile']);
+const CMD_LOCK_FILE = './.commandpidfile';
+const STDIO_PIPE_PATH = xpipe.eq('./.commandstdio');
 
 var timeouts = mage.core.config.get(['daemon', 'timeouts']) || {};
 
+function disableDaemonizedStdio(isCommandAppAlive) {
+	if (!isCommandAppAlive) {
+		return;
+	}
+
+	// Send stdio to a black hole
+	function disable(io) {
+		io.write = () => true;
+	}
+
+	disable(process.stdout);
+	disable(process.stderr);
+}
+
+function setDaemonizedStdio(isCommandAppAlive) {
+	if (!isCommandAppAlive) {
+		return;
+	}
+
+	// Forward logs to the command process
+	const conn = net.connect(STDIO_PIPE_PATH);
+
+	function enable(io) {
+		io.write = conn.write.bind(conn);
+	}
+
+	enable(process.stdout);
+	enable(process.stderr);
+}
 
 function out(str) {
 	process.stdout.write(str + '\n');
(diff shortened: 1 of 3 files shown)


stelcheck:
Closing, because I don't really care any longer.

stelcheck:
Closed the wrong issue... apologies for the confusion.

lib/daemon/index.js
```diff
@@ -1,22 +1,56 @@
...
 var timeouts = mage.core.config.get(['daemon', 'timeouts']) || {};
 
+function disableDaemonizedStdio(commandPid) {
+	if (!commandPid) {
+		return;
+	}
+
+	// Send stdio to a black hole
+	function disable(io) {
+		io.write = () => true;
```
ronkorving: The previous code was `() => false;`.
What changed about that?

stelcheck: `write` returns `true` in real life. I was fearing that `false` might somehow cause issues with dependencies.

ronkorving: Because of exit logs?

stelcheck: No, just because 'false' *may* be considered as an error. At any rate, returning `true` is more consistent.

lib/daemon/index.js
```diff
@@ -415,23 +460,20 @@ function removeSocketFiles() {
...
-				stdout,
-				stderr
-			} = process;
+	// Disable stdout/stderr if we are starting daemonized
+	processManager.once('started', () => disableDaemonizedStdio(commandPid));
 
-			stdout.write = () => false;
-			stderr.write = () => false;
-		}
+	// Upon exitting, check if we can forward exit logs
```
ronkorving: Very nitpicky, but suggestion when it comes to comment writing. I try to avoid "check if", because usually there's something we do, if ...

Compare to: `// Upon exiting, try to forward exit logs`
(Also, `exiting` has 1 t)

lib/daemon/index.js
stelcheck: According to the Node.js doc:

```
Returns: <boolean> false if the stream wishes for the calling code to wait for the 'drain' 
event to be emitted before continuing to write additional data; otherwise true.
```

Since we are *never* going to call drain, true is safer.

ronkorving: Ah yes, indeed 👍 
But friendly cc to @christopherobin who was involved in the original `return false` I think?

stelcheck: I was the one who wrote that for him, he just picked up the code from a separate branch I had created to solve some issues he was having with his initial PR.

stelcheck:
Fixed the comment.

lib/daemon/index.js
```diff
@@ -480,11 +522,15 @@ exports.init = function () {
 			}
 		}
 
+		var commandPid = getFilePid(CMD_LOCK_FILE);
```
ronkorving: Definitely not a blocker, but feel free to `const` all-the-things ;)

lib/daemon/index.js
```diff
@@ -1,22 +1,56 @@
...
+	function disable(io) {
+		io.write = () => true;
+	}
+
+	disable(process.stdout);
+	disable(process.stderr);
+}
+
+function setDaemonizedStdio(commandPid) {
+	if (!commandPid) {
```
ronkorving: The PID is not relevant to this function, so to pass it... is quite confusing (I know, because I've been trying to make sense of this PR in review ;))

Would it not make more sense to have a boolean returning function somewhere (instead of a PID, but by all means on the same basis), and if that boolean is false, to not *call* functions like (un)setDaemonizedStdio?

stelcheck: The issue is that you then end up having to carry a conditional *everywhere* (well. 4 places). 

I'll try to make this look a bit clearer.

lib/daemon/index.js
```diff
@@ -98,10 +132,27 @@ function destroyFilePid(filename) {
...
+function getCommandAppStatus() {
+	return !!getFilePid(CMD_LOCK_FILE);
+}
+
 
 function App(runstate, pid) {
+	// Force color output on logs - this is necessary so
+	// that the daemonized master process as well as its workers
+	// output colorized logs when running start/stop/restart/etc.
+	process.env.FORCE_COLOR = 1;
```
ronkorving: This probably has some drawbacks, no?

stelcheck: You are right. What we can do is have the daemonizer filter colors if it doesn't support them.

stelcheck: Just added the code to do that to this PR

PR #141, opened 2017-09-01
Development mode log

stelcheck:
Requires #115 
screenshot_2017-09-02_00-38-34

### DIFF
--- a/lib/mage/Mage.js
+++ b/lib/mage/Mage.js
@@ -1,19 +1,20 @@
 'use strict';
 
-var CODE_EXTENSIONS = [
+const CODE_EXTENSIONS = [
 	'.js',
 	'.ts'
 ];
 
-var EventEmitter = require('events').EventEmitter;
-var fs = require('fs');
-var util = require('util');
-var path = require('path');
-var async = require('async');
-var semver = require('semver');
+const EventEmitter = require('events').EventEmitter;
+const fs = require('fs');
+const util = require('util');
+const path = require('path');
+const async = require('async');
+const semver = require('semver');
+const chalk = require('chalk');
 
-var rootPath = process.cwd();
-var applicationModulesPath = path.resolve(rootPath, './lib/modules');
+const rootPath = process.cwd();
+const applicationModulesPath = path.resolve(rootPath, './lib/modules');
 
 
 function testEngineVersion(pkg, pkgPath, mage) {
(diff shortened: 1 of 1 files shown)


stelcheck:
Question: what should I do when the terminal width (number of columns) is insufficient?

christopherobin:
Above 45 characters and smaller than the current text:

```text
  _____             __  __           _      
 |  __ \           |  \/  |         | |     
 | |  | | _____   _| \  / | ___   __| | ___ 
 | |  | |/ _ \ \ / | |\/| |/ _ \ / _` |/ _ \
 | |__| |  __/\ V /| |  | | (_) | (_| |  __/
 |_____/ \___| \_/ |_|  |_|\___/ \__,_|\___|
```

Then below 45 characters do not show anything?

stelcheck:
OK, this should work fine.

screenshot_2017-09-04_12-58-58

screenshot_2017-09-04_12-59-17

lib/mage/Mage.js
```diff
@@ -706,16 +707,30 @@ Mage.prototype.start = function (cb) {
...
+	const log = mage.core.logger.warning;
+
+	if (process.stdout.columns > 45) {
+		log
+			.details(' _____             __  __           _')
+			.details('|  __ \\           |  \\/  |         | |')
+			.details('| |  | | _____   _| \\  / | ___   __| | ___')
+			.details('| |  | |/ _ \\ \\ / | |\\/| |/ _ \\ / _` |/ _ \\')
+			.details('| |__| |  __/\\ \V /| |  | | (_) | (_| |  __/')
+			.details('|_____/ \\___| \\_/ |_|  |_|\\___/ \\__,_|\\___|');
```
ronkorving: Have you considered writing straight to stdout instead? Especially since you're checking for terminal width, and the logger could be writing anywhere (syslog, graylog, you name it).

lib/mage/Mage.js
```diff
@@ -706,16 +707,30 @@ Mage.prototype.start = function (cb) {
 	return this;
 };
 
+function showDevelopmentModeWarning(mage) {
+	const log = mage.core.logger.warning;
+
+	if (process.stdout.columns > 45) {
```
ronkorving: I believe the terminal logger always writes to `stderr`, not `stdout`.

lib/mage/Mage.js
stelcheck: Is there a way to target a specific log writer? I would like to avoid having to colorize this myself...

stelcheck: Hmm, nevermind...

stelcheck:
screenshot_2017-09-11_15-09-03

Added some line spaces just to make it a bit more readable.

lib/mage/Mage.js
```diff
@@ -725,11 +742,13 @@ Mage.prototype.boot = function (cb) {
 		}
 
 		var keys = Object.keys(apps);
-		keys.forEach(function (appName) {
-			that.emit('setup.' + appName, apps[appName]);
-		});
+		keys.forEach((appName) => this.emit('setup.' + appName, apps[appName]));
```
ronkorving: If you're gonna rewrite this, I think a for-of loop would've been more readable. But no blocker.

PR #143, opened 2017-09-02
encoder: throw error if type is not a string

stelcheck:
The default encoders will now throw an error if the submitted
type is not a string or a Buffer. This should avoid situations
like this:

```javascript
state.archivist.set('foobar', {}, [], 'application/json', 'utf8');
```

Which would end up always returning an empty string, because:

  - new Buffer([]) is an empty string
  - [].toString() is also an empty string

### DIFF
--- a/lib/archivist/encoders.js
+++ b/lib/archivist/encoders.js
@@ -1,3 +1,5 @@
+'use strict';
+
 // encoding types:
 //   live (native JS objects, does not include Buffer instances)
 //   utf8 (string)
@@ -13,6 +15,15 @@
 var commonEncoders = {};
 var mediaTypeEncoders = {};
 
+class ArchivistEncodingError extends Error {
+	constructor(message, from, to, data) {
+		super(message);
+		this.name = 'ArchivistEncodingError';
+		this.from = from;
+		this.to = to;
+		this.data = data;
+	}
+}
 
 exports.registerCommonEncoder = function (fromEncoding, toEncoding, fn) {
 	if (!fn) {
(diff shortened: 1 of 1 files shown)


lib/archivist/encoders.js
```diff
@@ -21,7 +32,13 @@ exports.registerCommonEncoder = function (fromEncoding, toEncoding, fn) {
 
 	var name = fromEncoding + '-' + toEncoding;
 
-	commonEncoders[name] = fn;
+	commonEncoders[name] = function (data) {
```
christopherobin: Can't data be an actual object if the `fromEncoding` is `live` ?

stelcheck: From what I can tell, `registerCommonEncoder` is exposed but only called internally (somehow). And internally, all encoding rules are from string formats to buffers or the opposite. @ronkorving might be able to confirm my understanding of this bit of code.

ronkorving: Yeah.. I think it would be more appropriate to do this for each individual encoder function. That would also imply they will only have to test for *one* type, not both strings and buffers. The code below implies I can still pass the wrong type to an encoder (if it expects a buffer, I can send it a string).

stelcheck: From what I can tell, all encoders can take both string or Buffer:

```shell
> new Buffer('test')
<Buffer 74 65 73 74>
> new Buffer(new Buffer('test'))
<Buffer 74 65 73 74>
> 'asdas'.toString()
'asdas'
>
```

stelcheck: Ah, spoke too soon. A string `toString` is not working the same way as the `Buffer` object method.

stelcheck:
Code refactored according to previous comments.

PR #144, opened 2017-09-02
archivist: scan and per-query options

stelcheck:
- Added a scan method to archivist, and updated the documentation accordingly.
 -  Per-query option for mget. Query options will be merged with the options passed to the mget call.

### DIFF
--- a/docs-sources/includes/_Archivist.md
+++ b/docs-sources/includes/_Archivist.md
@@ -523,27 +523,27 @@ exports.getItemsForUser = function (state, userId, callback) {
   var topic = 'item';
   var partialIndex = { userId: userId };
 
-  state.archivist.list(topic, partialIndex, function (error, indexes) {
-    if (error) {
-      return callback(error);
-    }
-
-    var queries = indexes.map(function (index) {
-      return { topic: topic, index: index };
-    });
-
-    state.archivist.mget(queries, callback);
-  });
+  state.archivist.scan(topic, partialIndex, callback);
 };
 ```
 
-There are a few ways by which you can split and filter the data
-stored in your topics.
+With certain APIs you can provide only a portion of an index and search for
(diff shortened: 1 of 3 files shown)


stelcheck:
Updated PR to include the possibility to add per-query options.

docs-sources/includes/_Archivist.md
```diff
@@ -523,27 +523,27 @@ exports.getItemsForUser = function (state, userId, callback) {
...
-call to filter the list of keys to return. In the sample code here, we use this
-feature to return all items' full keys for a given user.
+There are a few ways by which you can split and filter the data
+stored in your topics:
+
+  1. You can use [archivist.list](https://mage.github.io/mage/api/classes/archivist.html#list) to list
+     all the indexes matching a given partial index
+  2. You can use [archivist.mget](https://mage.github.io/mage/api/classes/archivist.html#mget) to fetch
+     multiple indexes at once
+  3. You can use [archivist.scan](https://mage.github.io/mage/api/classes/archivist.html#list), which combines
```
ronkorving: Wrong anchor.

lib/archivist/index.js
```diff
@@ -526,7 +526,15 @@ function mgetAggregate(queries, retriever, cb) {
...
 			Object.keys(queries),
 			function (queryId, callback) {
-				retriever(queries[queryId], function (error, data) {
+				const query = queries[queryId];
+
+				if (!query.options) {
+					query.options = {};
+				}
+
+				query.options = Object.assign({}, options, query.options);
```
ronkorving: Rather than mutating the `query` object (which archivist does not own), why not do this instead:

```js
const queryOptions = query.options ? Object.assign({}, options, query.options) : options;

retriever(query, queryOptions, function (error, data) {
```

PR #145, opened 2017-09-02
batch/messageHooks: state metadata

stelcheck:
The batching system creates its own `state` independent from the state
created by each user commands. However, it sets no data on `state.data`,
whereas each user commands sets `state.data` to the value of the
metadata received from the batch executor.

This results with `messageHooks` functions receiving different metadata
than what user commands receive.

### DIFF
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -671,6 +671,10 @@ CommandCenter.prototype.executeBatch = function (acceptedEncodings, batch, query
 
 	state.appName = this.app.name;
 
+	if (metaData) {
+		state.data = metaData;
+	}
+
 	exports.emit('openPostConnection', this.app);
 
 	function cb(error, content, options) {


lib/commandCenter/commandCenter.js
```diff
@@ -670,6 +670,7 @@ CommandCenter.prototype.executeBatch = function (acceptedEncodings, batch, query
 	var state = new State();
 
 	state.appName = this.app.name;
+	state.data = metaData;
```
ronkorving: If no meta data is provided, will `metaData` still be an empty object? `state.data` defaults to an empty object, and I'm guessing we wouldn't want to overwrite that with `null` or `undefined`. At the user command level, there's an if-statement around the assignment:

https://github.com/stelcheck/mage/blob/a8a0faa1d74e3596ccf2316a195e781ec866f16c/lib/commandCenter/commandCenter.js#L495-L497

stelcheck:
Fixed issue as requested.

PR #147, opened 2017-09-02
JSON5 configuration format

stelcheck:
Adding support for JSON5 as a configuration format.

### DIFF
--- a/bin/mage
+++ b/bin/mage
@@ -1,20 +1,41 @@
 #!/usr/bin/env node
 
+const prettyjson = require('prettyjson');
+const chalk = require('chalk');
 const path = require('path');
 const cwd = process.cwd();
 
 let app;
 
+function logError(line) {
+	console.error(chalk.red.bold(line));
+}
+
+function logInfo(line) {
+	console.error(chalk.green(line));
+}
+
 try {
 	app = require(cwd);
 } catch (error) {
-	console.error('');
-	console.error('Failed to load MAGE application:');
-	console.error('');
-	console.error(error.stack);
-	console.error('');
-	console.error('Make sure to properly set the value of "main" in your package.json,');
-	console.error('and that your code is valid.');
+	const errorData = Object.assign({
+		stack: error.stack
+	}, error);
+
+	logError('');
+	logError('Failed to load MAGE application');
+
+	if (process.platform === 'win32') {
+		logError('-------------------------------');
+	} else {
+		logError('▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔');
+	}
+
+	console.error(prettyjson.render(errorData, null, 2));
+
+	logError('');
+	logError('Make sure to properly set the value of "main" in your package.json,');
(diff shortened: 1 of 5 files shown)


lib/config/config.js
```diff
@@ -40,7 +43,7 @@ function loadConfigFile(configPath) {
 		return require(configPath);
 	}
 
-	if (extension === '.json') {
+	if (extension === '.json' || extension === '.json5') {
```
stelcheck: One issue here: in case of syntax errors, the linter code below will not work properly against JSON5 files.

docs-sources/includes/_Configuration.md
```diff
@@ -3,11 +3,12 @@
 ## Format
 
 The configuration for your game is allowed to exist in either the
```
ronkorving: The word `either` implies 2 I believe, so this sentence maybe deserves rephrasing.

stelcheck:
Refactored that code a bit to make error management easier, and updated `bin/mage` to pretty-print error objects in a nicer way.

bin/mage
```diff
@@ -1,20 +1,41 @@
...
+	}, error);
+
+	logError('');
+	logError('Failed to load MAGE application');
+
+	if (process.platform === 'win32') {
+		logError('-------------------------------');
+	} else {
+		logError('▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔');
+	}
```
ronkorving: Really? :D

stelcheck: It's prettier 🌈

lib/config/config.js
```diff
@@ -29,6 +32,35 @@ var isInitialized = false;
 loggingService.addWriter('terminal', ['>=notice'], {});
 
 /**
+ * ConfigurationError class
+ */
+class ConfigurationError extends Error {
+	constructor(message, filename, error) {
+		super(message);
+
+		this.name = 'ConfigurationError';
```
ronkorving: Does extending `Error` not do this for you?

stelcheck:
```shell
> class H extends Error{}
[Function: H]
> (new H('asd')).name
'Error'
```

lib/config/config.js
```diff
@@ -29,6 +32,35 @@ var isInitialized = false;
...
+		this.filename = filename;
```
ronkorving: JS `Error` class [may already](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) expose a `fileName` property. To avoid confusion, you may want to rename this.

lib/config/config.js
ronkorving: I could've tested that. Sorry for the noise.

lib/config/config.js
stelcheck: Any suggestions?

stelcheck: I will merge it as-is for now; this is a Mozilla-specific error field, and I don't think it is something we can expect to see change anytime soon. And I fear using a different name may not be clear enough.

PR #151, opened 2017-09-04
processManager: mage worker id and parallelism fix

stelcheck:
Parallelism for worker restart was set to 4, and never changed;
instead, we now tell to start all workers at once on start, and
to reload them one by one on reload.

Unique ID per worker; if a worker dies and is replaced,
the replaced worker will have the same `mage.workerId`,
thus allowing for more consistent logging and metrics.

### DIFF
--- a/lib/mage/Mage.js
+++ b/lib/mage/Mage.js
@@ -41,6 +41,7 @@ function testEngineVersion(pkg, pkgPath, mage) {
 function Mage(mageRootModule) {
 	EventEmitter.call(this);
 
+	this.workerId = require('./worker').getId();
 	this.task = null;
 
 	this._runState = 'init';
--- a/lib/mage/worker.js
+++ b/lib/mage/worker.js
@@ -1,0 +1,11 @@
+'user strict';
+
+exports.getId = function () {
+	const id = process.env.MAGE_WORKER_ID;
+
+	if (!id) {
+		return false;
+	}
+
+	return parseInt(id, 10);
+};
--- a/lib/processManager/processManager.js
+++ b/lib/processManager/processManager.js
@@ -14,6 +14,12 @@ var numberOfWorkers = false;
 
 var realStartTime = Date.now();
 
+var workersIncrement = 0;
+
+function generateWorkerId() {
+	return workersIncrement += 1;
+}
+
 // expose information
 
 processManager.isMaster = false;  // true if the process is the master of the cluster (false for single-node mode)
@@ -73,7 +79,6 @@ function WorkerManager(initialMax) {
 
 	this.max = initialMax;
 	this.maxStartupTime = WORKER_STARTUP_TIMEOUT;
-	this.parallelism = 4;
 
 	// worker lists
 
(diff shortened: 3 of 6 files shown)


mage.ts
```diff
@@ -1286,6 +1286,17 @@ declare class Mage extends NodeJS.EventEmitter {
...
     /**
+     * The worker ID of the currently worker process
+     *
+     * Will return false if the process is the master process, or
+     * if MAGE is running in single mode.
+     *
+     * This differs from cluster.worker.id because it will remain consistent
+     * if restarted or reloaded.
+     */
+    workerId: number | false;
```
ronkorving: Would `undefined` not make more sense? If we can easily avoid mixing types (I don't consider `undefined` a type), I think that's preferable.

mage.ts
```diff
@@ -1286,6 +1286,17 @@ declare class Mage extends NodeJS.EventEmitter {
     isCodeFileExtension(ext: string): boolean;
 
     /**
+     * The worker ID of the currently worker process
```
ronkorving: `current`

lib/processManager/processManager.js
```diff
@@ -599,7 +611,8 @@ processManager.start = function (cb) {
 
 processManager.reload = function (cb) {
 	if (workerManager) {
-		workerManager.recycle(cb);
+		// Recycle one worker at a time
```
ronkorving: Why one at a time?

mage.ts
stelcheck: Agreed. I'll update.

lib/processManager/processManager.js
stelcheck: Right now, we kill the processes after we have reloaded. But if we want to have a second process to have the same "id", we should probably kill first, then restart, which means we should probably not reload all at once.

However, if this is not desirable, we'll still need to decide what we want in terms of reload logic; right now, it is hard-coded to always reload 4 at a time, which I don't think is a constant that necessarily make sense.

stelcheck: What should I do with this?

lib/mage/Mage.js
```diff
@@ -39,6 +39,9 @@ function testEngineVersion(pkg, pkgPath, mage) {
  */
 
 function Mage(mageRootModule) {
+	// Set up the worker id, if necessary
```
ronkorving: I don't see any if-statements, so not entirely sure what "if necessary" is talking about?

lib/mage/Mage.js
```diff
@@ -39,6 +39,9 @@ function testEngineVersion(pkg, pkgPath, mage) {
...
+	this.workerId = require('./worker').getId();
```
ronkorving: Maybe do this after the `EventEmitter.call(this);`?

lib/mage/Mage.js
stelcheck: https://github.com/mage/mage/pull/151/files#diff-490b45026160e1275d20c3b8f66ab6b4R3

The function itself has conditional logic in it.

lib/mage/Mage.js
stelcheck: Done.

lib/mage/Mage.js
stelcheck: Removed the comment overall.

stelcheck:
https://github.com/mage/mage/pull/151#discussion_r153061441 Still needs a response.

lib/mage/worker.js
```diff
@@ -0,0 +1,11 @@
+'user strict';
```
ronkorving: `use`

ronkorving: Oops, merged and forgot I left this comment :)

PR #152, opened 2017-09-06
mmrp: relay map corruption

stelcheck:
### Description 

MMRP relies on service discovery to keep its list of known relays
up to date. However, under certain conditions, and depending on the
behavior of the service discovery backend in use, you may end up
with two relays sharing the same URI. This, in and of itself, appeared
to be the cause of multiple issues.

You may also end up with case where lingering nodes are cleaned up
after the replacing node has been added; in such cases, a disconnect was
still triggered, thus leaving the MMRP cluster in an inconsistent state
where messages were silently black-holed.

### Encountered cases

Currently, only when using the ZooKeeper service discovery vault backend. This is because if a server unexpectedly crashes or is terminated with `SIGKILL`, the ZooKeeper node remains in the node tree until it expires.

However, I suspect other service discovery vault backends could potentially cause similar issues - to a certain degree.

### Solution

  1. On relay down, disconnect based on the cluster ID instead
     of disconnecting based on the URI (we still verify that the URI
     is consistent)
  2. On relay up, disconnect any relay with the same URI but a different
     clusterId

### DIFF
--- a/lib/msgServer/index.js
+++ b/lib/msgServer/index.js
@@ -587,19 +587,7 @@ function setupDiscovery() {
 			return;
 		}
 
-/*  // WHY IS THIS COMMENTED OUT? I HAVE NO IDEA!
-		// ignore clusterId-less peers
-
-		if (!data.clusterId) {
-			logger.debug.data({
-				peer: data,
-				us: ourMetadata
-			}).log('Ignoring service-down of service without clusterId:', announced.host + ':' + announced.port);
-			return;
-		}
-*/
-
-		mmrpNode.relayDown(uri);
+		mmrpNode.relayDown(uri, data.clusterId);
 	});
 
 	// announce our relay
--- a/lib/msgServer/index.js
+++ b/lib/msgServer/index.js
@@ -450,7 +450,8 @@ function setupDiscovery() {
 	var ourMetadata = {
 		game: cfgServiceName || mage.rootPackage.name,
 		version: mage.rootPackage.version,
-		clusterId: mmrpNode.clusterId
+		clusterId: mmrpNode.clusterId,
+		timestamp: Date.now()
 	};
 
 	service = serviceDiscovery.createService('mmrp', 'tcp');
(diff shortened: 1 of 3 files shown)


lib/msgServer/index.js
```diff
@@ -587,19 +587,7 @@ function setupDiscovery() {
 			return;
 		}
 
-/*  // WHY IS THIS COMMENTED OUT? I HAVE NO IDEA!
```
stelcheck: Removed this because it has been here forever.

stelcheck:
There appear to still be some cases this PR does not quite solve. I will continue to investigate.

lib/msgServer/mmrp/index.js
```diff
@@ -130,63 +135,107 @@ exports.MmrpNode = MmrpNode;
...
+		this.relays[clusterId] = new RelayConnection(uri, route, timestamp);
 	}
 
-	logger.verbose(this.clusterId, 'dealer connecting to', uri, '(cluster: ' + clusterId + ')');
+	// stelcheck (mt): I do not know why, but the setTimeout appears to solve
+	// an issue with reconnection on restart-after-crash if the
+	// nodes are emitter in non-chronological order
 
-	this.dealer.connect(uri);
+	setTimeout(() => {
```
stelcheck: I have no idea why this is needed cc @ronkorving @christopherobin

stelcheck:
All issues concerning this PR should now have been fixed, but I wish to leave this PR sitting here until I have seen it work consistently on some test deployments I manage. Will remove the DNM notice then. 

In the meantime, please feel free to review.

stelcheck:
@christopherobin could you have a look at this one please?

ronkorving:
I don't see anything blatantly wrong, so in that sense LGTM. But this needs testing more than anything. Happy to see this merged as-is if it'll be exposed to some real world environment before release.

stelcheck:
This actually solves an issue on my current project, so I can tell it does what it intends to do. As far as not introducing new bugs, I guess its a bit trickier.

What we could do is merge and schedule for a 1.4 release. Then projects running off master would be able to get some runtime to confirm all is good.

ronkorving:
I'm fine if @christopherobin is fine 👍

ronkorving:
Decided with @stelcheck triggered by feedback from @christopherobin that we will merge this now.

PR #156, opened 2017-09-11
dead link checker

stelcheck:
Also updated the package-lock.json file since it seems like
it was pretty much out of date.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "test:outdated": "npm outdated --color always --silent || true",
     "test:security": "retire && npm run test:outdated",
     "test:lint": "eslint --ext .js,.json .",
+    "test:docs": "blcl ./docs -ro",
     "test:code-analysis": "run-s test:lint test:security",
     "test:unit": "mocha ./test/index.js",
     "test:unit:filter": "npm run test:unit -- --grep",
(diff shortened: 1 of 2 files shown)


ronkorving:
Maybe we can add mdlint to this?

stelcheck:
We could, but I would perhaps indeed let this go into another PR.

stelcheck:
PR was updated to solve merge conflicts.

package.json
```diff
@@ -18,6 +18,7 @@
     "test:outdated": "npm outdated --color always --silent || true",
     "test:security": "retire && npm run test:outdated",
     "test:lint": "eslint --ext .js,.json .",
+    "test:docs": "blcl ./docs -ro",
```
ronkorving: We don't run this on `npm test`?

stelcheck: We can't. Reason is that sometimes you want to link to, say, the API doc, but it hasn't been published yet.

ronkorving: Maybe run it in a prepublish hook or something?

stelcheck: We could... just not sure if there might be any other caveats I am not thinking of. 

I'd say we could add this to pre-publish at a later time, once we've used that tool a bit more by hand.

PR #160, opened 2017-09-12
topics: json as default media type

stelcheck:
Since Tomes is not widely used (or currently documented, for that
matter), making the default 'application/json' is more comprehensive.

### DIFF
--- a/lib/archivist/configuration.js
+++ b/lib/archivist/configuration.js
@@ -427,7 +427,7 @@ function registerTopicConfig(topic, cfg) {
 
 	// initialize topicConfig with defaults
 
-	var defaultMediaTypes = ['application/x-tome', 'application/octet-stream'];
+	var defaultMediaTypes = ['application/json'];
 
 	var topicConfig = {
 		readOptions: {


AlmirKadric:
hmmm that is a very fundamental change and I'm not sure its a good one. In most instance till date, tomes has always been the main form of document type, very rarely did we use JSON. Even with recent changes there is no official replacement to tomes and how we manage data synchronisation.

This would mean additional boiler plate (not a lot if factory function are used for topics). However any guides or training material will need to be updated to explain the reasoning for media-types. I have to say I disagree with this change at this time (until an official tom replacement is in place).

This being said not a blocker if everyone else understands the above concerns and is still in favor of the requested change.

ronkorving:
What I've concluded at Wizcorp recently, is that Tomes has lost its status as the official default (to most of the staff). Of course if people want to use it, I won't stop them. We all know the ramifications at this point. But, I think the number of people who *want* to use Tomes atm are probably fewer than the people who *do* want to use it.

I'm personally also kinda done with Tomes, and while there is no official replacement, I don't think we need one. Around JSON you can choose a lot of technologies. Tomes *is* one technology. Recently I've used Redux and I'm quite happy with the result. I will *not* say "let's make Redux the default", but I *can* use Redux around JSON, not around Tomes.

So a default that leaves the door open to many choices, to me makes more sense than a default that can only be used one way.

AlmirKadric:
Yes but until we have an official replacement we cannot strip Mage of this feature. This is still one of the biggest selling points of Mage. And whilst people dislike tomes, if asked to implement a replacement on their own versus using it with all its faults, a part from some our most seasoned Mage users, most will use safe default.

Also this has ramifications on training. Now training has to teach how someone can implement a replacement to tomes (once we officially remove it from the feature set).

Whilst I understand you're frustrations, we need to provide alternatives before removing a feature like this (even if it is to change it away from the default for now).

I just want to make sure people understand the many facets of this choice before it is made.

christopherobin:
It doesn't remove Tomes from deps though (which I'm fine with) and it's still relatively easy to use them in an application.

I agree with the move to `application/json` but I feel that to do this move the documentation requires a whole new section about data synchronization before it should be done.

That section should explain how to use either Tomes or events (since both are still bundled with MAGE), with pros and cons of both solutions.

stelcheck:
> That section should explain how to use either Tomes or events 
> (since both are still bundled with MAGE), with pros and cons of both solutions.

https://mage.github.io/mage/#events describes how events can be emitted. We could work on improving this section to make it a bit more comprehensive.

Tomes is currently undocumented because I was under the impression, from general feedback, that it was essentially on its way out, in favor of simply emitting updates through `state.emit`; I have heard very few (if any) positive use cases for Tomes, and to me, it wouldn't make sense to keep in a system nobody really wants to use long term. But of course, this situation can be rectified if needed.

AlmirKadric:
From what I have seen, I would say rather than tomes being on its way out, people have parts of it they dislike and would like to see it fixed (which would essentially mean a rewrite due to fundamental architecture flaws within MMRP as well as Tomes), thus why it seems like its on its way out. However it needs a replacement as the problem it solves is still a legitimate one. In @ronkorving case he used Redux to fill this void (some MMRP issues still remain, mainly the usercommand vs message stream race condition).

I am not against any of these changes, but we need to make sure MAGE has this kind of support as a core feature. This is why I believe we need to officiate a replacement before we scrap it.

stelcheck:
> which would essentially mean a rewrite due to fundamental architecture 
> flaws within MMRP as well as Tomes

The issues with MMRP are currently being fixed (see #152 for instance) and have not seen any unsolvable architectural issues so far. If you do know of fundamental architectural issues, please share them with me in private chat (so not to derail the current discussion from its main topic).

I'll try to get more people to join here and provide feedback.

ronkorving:
@AlmirKadric 
> which would essentially mean a rewrite due to fundamental architecture flaws within MMRP as well as Tomes

I think a rewrite of Tomes to fix what is wrong with it, would mean it would no longer be Tomes. I have little reason to believe it can all be fixed without having to change its API. Tomes solve a problem, but that doesn't make everything that solves the same problem "Tomes" :)

I think MMRP has no relevance in this topic. Message stream does, as message stream and user command instant responses *can* indeed cause race conditions, and Tomes do suffer from that. But so would any other data synchronization mechanisms... to some extent anyway. But since none of this is Tomes-specific, I don't see that as very relevant to this topic.

> I am not against any of these changes, but we need to make sure MAGE has this kind of support as a core feature. This is why I believe we need to officiate a replacement before we scrap it.

Technically, event emission alone could be a fair replacement. The way I used Redux is just a fancy version of that. But really, in essence, it's just server-side code saying "this changed", and the client saying "got it, let me update the data I have". And unlike Tomes, the client-side data doesn't even have to follow the server-side schema for that to work. So it's quite a free system, and it works.

I appreciate the need to perhaps dedicate a chapter in the documentation on data-synchronization, and the approaches we have so far used with some success.

AlmirKadric:
@ronkorving I think you missed my point. I completely agree with you on all those points. My point is let's not remove/destroy and fix later, but rather fix first then replace. My concern is the removal or altering of default behaviour without having an alternative prepared is a bad experience to give to the users. Whilst event emission might be a sufficient replacement on a technical level, we should a the very least properly document how it can be used as such or any other alternatives before this change is approved and further to this make alternatives readily available before removing tomes all together (I know this is a separate discussion, but that is the ultimate intention of the PR).

Whilst event emission might be enough for you if new users don't understand how to use what's the point? Also whilst it might be enough for you, what about all other users?

stelcheck:
Following the conversation above, two questions were made clear; I went around and asked these questions to six developers with extensive experience with MAGE. 

  1. Is Tomes a sane, desirable default (1 yes, 4 no and 1 abstention)
  2. Is Tomes, or a Tome-like solution, a necessity for providing a sane game development ecosystem (2 yes, 3 no and 1 abstention).

I also took the time of making a broader qualitative inquiry with a larger pool of developers to try to pinpoint common opinions about Tomes.

  1. Almost unanimously, developers who have experienced Tome only through a training program find it hard to understand and buggy, but find it nice to use once they get over the learning curve.

  2. Almost unanimously, developers who have worked with Tomes on released games don't see the value of it, citing frequent bugs and usability issues

  3. Feelings are more mixed about whether `state.emit` is a sufficient replacement to Tomes; while some feel it is, others feel it would leave less seasoned developers out to dry and figure out how to do synchronization themselves. Some also have mentioned that smaller games would likely be easier to make using Tomes.

  4. Most developers feel Tomes should be out of MAGE and into its own external module. 

I will be leaving this PR open a bit longer, just in case anyone feels additional comments are warranted. However, I would kindly request that additional comments remain on-topic and that they avoid repeating what was previously mentioned here; if you wish to voice your support for a previous comment, please add a Reaction on the comment instead.

ronkorving:
@AlmirKadric I understand where you're coming from.

My suggestion for documentation and training would be this:

- Teach that MAGE has an event system to communicate with the client and how it can be used (through examples).
- Teach that it can be used with Tomes to synchronize data to the client (how to configure, and how to use Tomes).

This order of teaching, I think, would avoid a certain case I saw in the past and was rather stunned by: people knew about tomes, but not about events. In that scenario, Tomes remain very magical (perhaps fine at first) because what people understand as "Tomes" is *really* "events + Tomes". So teaching first about the event system (in very simple terms) would remove some magic there, which I think would help people understand 2 things better:

1. How Tomes does what it does
2. Tomes is not the only way to communicate with the client (when all you know is Tomes, every problem starts to look like it needs a tome-solution)

stelcheck:
First, thank you to everyone who weighed in here. I really appreciate all the feedback we were able to gather.

I recently had to play with Tomes more closely, see how developers can struggle with them, and so on. In fact, I even added tomes topics support to https://github.com/mage/mage-validator, which deals with some of the API issues.

But it is still quite a magical tool to play with, one that does a *lot* in the background. And black magic is never good for newcomers. 

Thus, I believe that tomes, *while very useful in some cases*, isn't a sane default that will help make MAGE more accessible. If anything, it is more valuable to learn raw MAGE *first*, then tackle Tomes as a more advanced topic. 

In that spirit, I will be merging this PR.

PR #164, opened 2017-09-14
Standard MageError class for user-land

### LINKED ISSUE #135: RFC: MageError class(es)
In async user commands, throw errors will either return `error.code` or `error.message`. But when I throw right now, I often do:

```javascript
throw new Error('Invalid request, partner set deck')
```

We should probably promote something like:

```javascript
throw new MageError({
  code: SomeEnumOrMap.ErrorCode,
  message: 'Something bery bery baaad happened',
  details: {
    some,
    request,
    data
  }
})
```

This would:

  1. Encourage developers to set explicit error codes which can then be shared with the client (and which the client should be able to react to)
  2. Still log the message
  3. Attach error-related data with the error instead of having to log those separately
  4. Provide a base class to extend:

> lib/consts/errors.ts

```typescript
import { MageError } from 'mage'

// throw new QueryError(typedDetails)
export QueryError extends MageError<IQueryDetails> {
  public code = 'query_error'
  public message = 'Query failed unexpectedly'
}
```

stelcheck:
This should help:

  1. Standardize error management
  2. Allow for throwing errors (and interrupt the execution flow)
     without necessarily logging an error
  3. Allow to attach additional details to the error to be logged, thus
     removing the need for an additional log call prior to throwing
  4. Attach a response code to be sent to the client; this also has the
     advantage of being part of the log data, thus making it easier
     to search for error occurrences by code instead of by message
     content.

Requires #112 
Fixes #135

### DIFF
--- a/docs-sources/includes/_Modules.md
+++ b/docs-sources/includes/_Modules.md
@@ -308,3 +308,68 @@ instead of relying on MAGE to serialize the return data. This can be useful when
 will serve certain pieces of data which will not change frequently. When you wish to do so, simply make
 sure that `exports.serialize` is set to `false`, and to manually return a stringified version of your
 data.
+
+## Error management
+
+> lib/modules/players/usercommands/register.js
+
+```javascript
+'use strict';
+
+const {
+  players,
+  MageError
+} = require('mage');
+
+class NoHomersError extends MageError {
+  constructor(data) {
+		super(data);
+
+    // Code to return to the client (default: server)
+    this.code = 'server';
+
+    // Log level to use to log this error (default: error)
+    this.level = 'warning';
+
+    // Details to log alongside the message (default: undefined)
+    this.details = data.details;
+
+    // Error type - traditionally the class name
+ 		this.type = 'NoHomersError';
+ 	}
+}
+
+module.exports = {
+  acl: ['*'],
+  async execute(state, username, password) {
+    if (username.length > 10) {
+      // Using MageError directly
+      throw new MageError({
+        code: 'username_length_exceeded',
+        level: 'warning',
+        message: 'Username is more than 10 character',
+        details: {
+          received: username
+        }
+      })
+    }
+
+    if (username === 'homer') {
+      // Using a custom error class inheriting MageError
+      throw new NoHomersError({
+        message: 'We already have one Homer anyway '
+      })
+    }
+
+    return await players.register(state, username, password);
+  }
+};
+```
+
+When using `async/await`, you will want to customise the behavior of your errors
+in regards to how they will be logged, what data will be logged, what message will be
+returned to the user, and so on.
+
+To do so, MAGE provides the `MageError` error class, which can be used as-is or
+extended as needed. This error class will allow you to control precisely what you will
+log in case of error, and what should be returned to the user.
--- a/lib/mage/MageError.js
+++ b/lib/mage/MageError.js
@@ -1,0 +1,23 @@
+'use strict';
+
+const mage = require('lib/mage');
+const loggingService = require('lib/loggingService');
+const logLevels = loggingService.getAllChannelNames();
+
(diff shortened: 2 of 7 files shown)


stelcheck:
Adding the label WIP just because documentation would be needed before merging.

stelcheck:
Documentation was completed, ready for review!

lib/mage/MageError.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+module.exports = class MageError extends Error {
+	constructor(data) {
+		super(data.message || 'Mage Error');
+
+		this.code = data.code || 'server';
+		this.level = data.level || 'error';
```
ronkorving: If the level is unrecognized, should we default it to `alert` or something, so that it can be fixed?

Throwing or otherwise seems unwise, since the user is trying to log something here, and we shouldn't get in the way of that. But especially because of that, it does make sense to detect it, fix it on the spot, and escalate it.

docs-sources/includes/_Modules.md
```diff
@@ -308,3 +308,68 @@ instead of relying on MAGE to serialize the return data. This can be useful when
...
+		super(data);
+
+    // Code to return to the client (default: server)
+    this.code = 'server';
+
+    // Log level to use to log this error (default: error)
+    this.level = 'warning';
+
+    // Details to log alongside the message (default: undefined)
+    this.details = data.details;
```
ronkorving: `details` in the logger used to be like a "long message". Here I guess it's a dictionary (from looking at the code example below)?

lib/mage/MageError.ts
```diff
@@ -0,0 +1,53 @@
...
+     * The log level to use when logging this error (default: error)
+     *
+     * Some types of errors are perfectly normal; for instance,
+     * a user failing to authenticate may throw an error, but
+     * you might not want to have it logged as an error.
+     *
+     * Setting the level will allow you to decide whether this
+     * error should be ignored, considered benign, or serious.
+     */
+    level?: 'debug' | 'warning' | 'error' | 'alert' | 'emergency',
```
ronkorving: Does debug have a place in this? Also, I think `critical` is missing.

lib/mage/MageError.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The error code
+     *
+     * This code will generally be the code that will be returned
+     * to the client if this error is thrown during a user command
+     * request.
+     */
+    code: string,
```
ronkorving: I believe `state` allows `code` to be both `string` or `number`.

lib/mage/MageError.ts
stelcheck: Who knows :D I would simply support all log levels just to make sure we have more choice.

lib/mage/MageError.ts
stelcheck: What happens if its a number?

stelcheck: I see the code is internally stringified.

Personally, I would prefer to be explicit at least in typescript about this - I'd hate to see confusion with a dev returning a number only to wonder why it's being received as a string on the client.

lib/mage/MageError.ts
stelcheck: I take that back. The idea behind debug being in the list was that some userland errors are perfectly natural, and you wouldn't necessarily want to log them in production but still make sure to return a specific error code (insufficient funds, etc)

lib/mage/MageError.js
stelcheck: Defaulting to alert but logging a warning about the discrepancy.

docs-sources/includes/_Modules.md
ronkorving: Fine as-is 👍

PR #166, opened 2017-09-15
mmrp: generate ID based on IP list

### LINKED ISSUE #153: zookeeper: nodes should be named based on uri, not clusterId
While working on https://github.com/mage/mage/pull/152 I realized that ZooKeeper could potentially leave lingering nodes behind. The aforementioned PR solves the main issue with that, however, the cluster information will still take up to 30 seconds to become consistent. This means that systems relying on service discovery information - such as https://www.npmjs.com/package/mage-module-shard - will route traffic based on an inconsistent view of the cluster.

Instead, what we should do is:

  1. Name ZooKeeper nodes using the URI, instead of the cluster ID. This should reduce the number of lingering nodes after a crash-and-restart
  2. Make sure that the zookeeper service discovery backend propagates correctly the overwrite. 

While this does not eliminate all lingering node cases, it should help reduce their number - essentially, the only time where nodes may be left to linger is if a cluster is downsized, and nodes are killed with `SIGKILL`.

@ronkorving @christopherobin Does that make sense to you guys?

stelcheck:
- Use the list of local IPs to determine a consistent cluster ID.
- Zookeeper: generate a consistent ID for storing entry into ZooKeeper

Requires #152 
Fixes #153

### DIFF
--- a/lib/msgServer/mmrp/index.js
+++ b/lib/msgServer/mmrp/index.js
@@ -7,10 +7,11 @@ const zmq = requirePeer('zmq');
 const util = require('util');
 const EventEmitter = require('events').EventEmitter;
 const Envelope = require('./Envelope.js');
-const mage = require('../../mage');
+const mage = require('lib/mage');
 const logger = mage.core.logger.context('mmrp');
 const createDealer = require('./createDealer.js');
 const createRouter = require('./createRouter.js');
+const serviceDiscovery = require('lib/serviceDiscovery');
 const slice = Array.prototype.slice;
 
 exports.zmqVersion = zmq.version;
@@ -20,6 +21,8 @@ exports.Envelope = Envelope;
 const SEND_RETRY_INTERVAL = 200;
 
 const inProcessDealerCount = {};
+
+const ENV_MAGE_MMRP_MASTER_PORT = 'MAGE_MMRP_MASTER_PORT';
 
 // Call the callback if defined
 const callMeMaybe = (cb, error) => cb && cb(error);
@@ -74,31 +77,18 @@ function ClientConnection(route) {
  * @constructor
  */
 
-function MmrpNode(role, cfg, clusterId) {
+function MmrpNode(role, cfg) {
 	EventEmitter.call(this);
 
 	this.isRelay = (role === 'relay' || role === 'both');
 	this.isClient = (role === 'client' || role === 'both');
 
 	assert(this.isRelay || this.isClient, 'MmrpNode must be relay, client or both');
-	assert(clusterId);
 
 	// keeping track of who we're connected to
 
 	this.relays = {};   // my peers, master-relay or both { identity: RelayConnection }
 	this.clients = {};  // my clients { identity: ClientConnection }
-
-	this.clusterId = clusterId;
-
-	// dealer
-
-	this.identity = this.isRelay ?
-		this.clusterId :
-		createDealerIdentity(this.clusterId);
-
-	this.dealer = createDealer(this.identity);
-
-	assert.equal(this.identity, this.dealer.getsockopt(zmq.ZMQ_IDENTITY));
 
 	// router
 
@@ -108,10 +98,32 @@ function MmrpNode(role, cfg, clusterId) {
 	if (this.isRelay) {
 		this.router = createRouter(cfg);
 		var routerUri = this.router.getsockopt(zmq.ZMQ_LAST_ENDPOINT);
+
 		if (routerUri) {
 			this.routerPort = parseInt(routerUri.substr(routerUri.lastIndexOf(':') + 1), 10);
+
+			// Share MMRP master port through the environment
+			process.env[ENV_MAGE_MMRP_MASTER_PORT] = this.routerPort;
+
+			// Generate the cluster ID based on the port the master listens on
+			this.clusterId = serviceDiscovery.helpers.generateId(this.routerPort);
+
+			// Set the identity of the router socket to the cluster ID
+			this.router.setsockopt(zmq.ZMQ_IDENTITY, new Buffer(this.clusterId));
 		}
+	} else {
+		// Retrieve the cluster ID from the environment
+		const masterPort = parseInt(process.env[ENV_MAGE_MMRP_MASTER_PORT], 10);
+
+		// Generate the cluster ID based on the port the master listens on
+		this.clusterId = serviceDiscovery.helpers.generateId(masterPort);
 	}
+
+	// dealer
+
+	this.identity = this.isRelay ? this.clusterId : createDealerIdentity(this.clusterId);
+	this.dealer = createDealer(this.identity);
+	assert.equal(this.identity, this.dealer.getsockopt(zmq.ZMQ_IDENTITY));
 
 	// mmrp socket event handling
 
--- a/lib/serviceDiscovery/helpers.js
+++ b/lib/serviceDiscovery/helpers.js
@@ -1,0 +1,43 @@
+'use strict';
+
+const os = require('os');
+const crypto = require('crypto');
+const memoizee = require('memoizee');
+
+const interfaces = os.networkInterfaces();
+const ifaceNames = Object.keys(interfaces);
+
+/**
+ * Builds a list of every announce-able IP on the server
+ *
+ * @returns {string} The ip list
+ * @todo Support from env variables?
+ */
+exports.getAnnounceIps = memoizee(function () {
+	const announceIps = [];
+
+	for (var i = 0; i < ifaceNames.length; i++) {
+		var addresses = interfaces[ifaceNames[i]];
+
+		for (var j = 0; j < addresses.length; j++) {
+			var address = addresses[j];
+
+			// skip anything that doesn't interest us
+			if (address.internal) {
+				continue;
+			}
+
+			announceIps.push(address.address);
+		}
+	}
+
+	return announceIps.sort();
+});
+
+exports.generateId = function (port) {
+	const generator = crypto.createHmac('sha256', 'secret');
+	exports.getAnnounceIps().forEach((ip) => generator.update(ip));
+
+	const id = generator.digest('hex').substring(0, 16) + ':' + port;
+	return id;
+};
--- a/lib/modules/session/index.js
+++ b/lib/modules/session/index.js
@@ -9,7 +9,6 @@ var timeCodeToSec = mage.core.helpers.timeCodeToSec;
 
 // config
 
-var clusterId = mage.core.msgServer.getClusterId();
 var defaultLanguage = 'en';
 var ttlCode = config.get(['module', 'session', 'ttl'], '10m');      // session timeout in time-code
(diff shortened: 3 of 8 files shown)


stelcheck:
Could we get this moving forward? Just realized I need this for a project I am working on.

AlmirKadric:
i reviewed this, and nothing really stands out
this being said i would need to really go deep into the system to properly review logical issues
very hard to do that from a diff
cc: @ronkorving

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+let announceIps;
+
+/**
+ * Builds a list of every announce-able ips on the server
+ *
+ * @returns {string} The ip list
+ * @todo Support from env variables?
+ */
+exports.getAnnounceIps = memoizee(function () {
+	if (announceIps) {
```
ronkorving: You're double memoizing :)
Once by doing it yourself, and once with memoizee.

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const os = require('os');
+const crypto = require('crypto');
+const memoizee = require('memoizee');
+
+const interfaces = os.networkInterfaces();
+const ifaceNames = Object.keys(interfaces);
+
+// choose the first valid ip as the one we will announce
```
ronkorving: I can't tell what this comment applies to... Not to the `let announceIps` below I assume.

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+			if (address.internal) {
+				continue;
+			}
+
+			// found it, store it and break
+			announceIps.push(address.address);
+		}
+	}
+
+	// just in case
```
ronkorving: How is this "just in case"?

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+
+		for (var j = 0; j < addresses.length; j++) {
+			var address = addresses[j];
+
+			// skip anything that doesn't interest us
+			if (address.internal) {
+				continue;
+			}
+
+			// found it, store it and break
```
ronkorving: There's no `break`

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+const memoizee = require('memoizee');
+
+const interfaces = os.networkInterfaces();
+const ifaceNames = Object.keys(interfaces);
+
+// choose the first valid ip as the one we will announce
+let announceIps;
+
+/**
+ * Builds a list of every announce-able ips on the server
```
ronkorving: English: "every announce-able IP" (not plural)

lib/serviceDiscovery/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+
+	// just in case
+	return announceIps;
+});
+
+exports.generateId = function (port) {
+	const generator = crypto.createHmac('sha256', 'secret');
+	exports.getAnnounceIps().forEach((ip) => generator.update(ip));
```
ronkorving: I understand you're trying to get this ID to be consistent for the machine this is running on, but since we're doing `Object.keys` on line 8 and not sorting, I don't think order can be guaranteed. It would probably be wise to add a `.sort()` on this line:

```js
exports.getAnnounceIps().slice().sort().forEach((ip) => generator.update(ip));
```

(turns that code a bit ugly now, but ... I hope it illustrates the point)

lib/msgServer/mmrp/index.js
```diff
@@ -103,11 +98,33 @@ function MmrpNode(role, cfg, clusterId) {
...
 			this.routerPort = parseInt(routerUri.substr(routerUri.lastIndexOf(':') + 1), 10);
+
+			// Share MMRP master port through the environment
+			process.env[ENV_MAGE_MMRP_MASTER_PORT] = this.routerPort;
+
+			// Generate the cluster ID based on the port the master listens on
+			this.clusterId = serviceDiscovery.helpers.generateId(this.routerPort);
+
+			// Set the identity of the router socket to the cluster ID
+			this.router.setsockopt(zmq.ZMQ_IDENTITY, new Buffer(this.clusterId));
```
ronkorving: I notice this wasn't done before for the router, was it? Any idea why and how does introducing it here change the behavior?

lib/msgServer/mmrp/index.js
```diff
@@ -103,11 +98,33 @@ function MmrpNode(role, cfg, clusterId) {
...
 		}
+	} else {
+		// Retrieve the cluster ID from the environment
+		const masterPort = process.env[ENV_MAGE_MMRP_MASTER_PORT];
```
ronkorving: Just for consistency, you may want to `parseInt` this.

lib/msgServer/mmrp/index.js
```diff
@@ -103,11 +98,33 @@ function MmrpNode(role, cfg, clusterId) {
...
+
+		// Generate the cluster ID based on the port the master listens on
+		this.clusterId = serviceDiscovery.helpers.generateId(masterPort);
 	}
 
+	// dealer
+
+	this.identity = this.isRelay ? this.clusterId : createDealerIdentity(this.clusterId);
```
ronkorving: I get the impression from looking at the code that `this.clusterId` could be undefined. That is, there may be code paths that lead to it not getting set.

lib/msgServer/mmrp/index.js
```diff
@@ -130,28 +147,64 @@ exports.MmrpNode = MmrpNode;
...
+	const relay = this.relays[clusterId];
 
 	// if we're already connected to this URI, we ignore this request
 
 	if (relay && relay.uri) {
-		if (cb) {
-			cb();
-		}
-		return;
+		return callMeMaybe(cb);
```
ronkorving: Haven't you ruled this out by the loop you introduce above? Either we've returned because the existing relay is newer, or we've deleted the existing one.

lib/msgServer/mmrp/index.js
```diff
@@ -130,28 +147,64 @@ exports.MmrpNode = MmrpNode;
...
+	const timestamp = data.timestamp;
+
 	// we cannot connect through routers (long routes), we are establishing a direct connection
 
 	assert.strictEqual(typeof uri, 'string');
 	assert.strictEqual(typeof clusterId, 'string');
 
-	var route = [clusterId];
-	var relay = this.relays[clusterId];
+	// check if other relay(s) with the same URI exist, and disconnect them;
```
ronkorving: Except.. you don't really disconnect them. I understand you don't really want to explain that weird disconnect situation twice, but .. yeah :-/

lib/msgServer/mmrp/index.js
```diff
@@ -130,28 +147,64 @@ exports.MmrpNode = MmrpNode;
...
+			return callMeMaybe(cb);
+		}
+
+		logger.verbose.data({
+			oldRelayId: relayClusterId,
+			newRelayId: clusterId
+		}).log(this.clusterId, 'found relay with same uri, removing the relay');
+
+		// We delete the relay reference, but do not disconnect; quick disconnect-reconnect
+		// appear to confuse ZeroMQ, and may permanently break the connection in some cases.
```
ronkorving: Some more information, perhaps documentation or a link to a gist, or .. you name it, would be really helpful for future generations.

lib/msgServer/mmrp/index.js
```diff
@@ -168,25 +221,29 @@ MmrpNode.prototype.connect = function (uri, clusterId, cb) {
...
 
-			logger.verbose(this.clusterId, 'dealer disconnecting from', uri, '(cluster: ' + clusterId + ')');
+	if (!relay) {
+		return verbose.log('dealer disconnect requested but relay does not exist');
+	}
 
-			this.dealer.disconnect(uri);
-			return;
-		}
+	if (uri !== relay.uri) {
```
ronkorving: How could this ever happen?

lib/serviceDiscovery/helpers.js
stelcheck: No idea. Guessing that was a comment from a prior prototype implementation I had... apologies for the noise.

lib/serviceDiscovery/helpers.js
stelcheck: Well... just in case you care? 

*sigh* guess that was a late night commit. Again, apologies for that.

lib/serviceDiscovery/helpers.js
stelcheck: Think I remember now: initially I wanted to use the first announce-able IP, but I didn't trust that this would be consistent enough on all platform and ended up using **all** announced IPs instead. Comments are from that initial concept.

lib/serviceDiscovery/helpers.js
stelcheck: Good catch. I'll sort in getAnnounceIps instead though.

lib/msgServer/mmrp/index.js
stelcheck: Been two months + so memory is fuzzy. But IIRC, in the past indentity was ephemeral, so setting an explicit identity didn't matter. Now that the ID is consistent across reboots, setting it is necessary to ensure queued messages will find their way to that node post-reboot.

lib/msgServer/mmrp/index.js
stelcheck: I think its undefined if you are in single mode (when zmq.ZMQ_LAST_ENDPOINT on line 100 returns nothing).

lib/msgServer/mmrp/index.js
stelcheck: Suggestions?

lib/msgServer/mmrp/index.js
stelcheck: I'd say its something we'd want to have blow up in our face if it *did* ever happen.

lib/msgServer/mmrp/index.js
stelcheck: I think that's incorrect:

1. We go through the list of nodes
2. We continue if this node is the node we are connecting to
3. We continue if the URIs do not match
4. If we find a more recent node, we return, since the node we are trying to connect is outdated
5. We disconnect any node reaching this point

Therefore, once we reach that point, it's entirely possible that we have disconnected a bunch of servers but were not connected ourselves.

stelcheck:
Have gone through most of the comments, fixed what I could and left follow up comments elsewhere, except for https://github.com/mage/mage/pull/166#discussion_r153679396 for which I'll need to go through the ZMQ mailing list (IIRC that's where I found some info on that subject).

lib/serviceDiscovery/helpers.js
stelcheck: Sorting there actually ends up having an adverse effect. See https://github.com/mage/mage/pulls for more details.

PR #168, opened 2017-10-18
couchbase helper: ignore auth errors on open

### LINKED ISSUE #167: Couchbase "waiting for SET" failure on 2.4.0
I noticed while trying to bump node-couchbase to 2.4.0 (2.3.7 wasn't building on Windows), archivist-create's "waiting for SET" was failing hard (error code 2). The first bucket was created, but because "waiting for SET" failed, I had to run through it multiple times to get both of my buckets to be created. On the second one, it also failed "waiting for SET", but again, the bucket did get created.

stelcheck:
An auth fail after a successful bucket creation would
mean that the credentials are valid, but that the
bucket has not fully been created yet.

Privatizing some methods which are not being used anywhere
else than internally.

Fixes #167

### DIFF
--- a/lib/archivist/vaults/couchbase/couchbaseHelpers.js
+++ b/lib/archivist/vaults/couchbase/couchbaseHelpers.js
@@ -43,12 +43,16 @@ function handleErrorPerErrorCode(error, exemptedCodes, callback) {
  * See https://github.com/couchbase/libcouchbase/blob/master/include/libcouchbase/error.h
  * for more details about the meaning of each error codes.
  *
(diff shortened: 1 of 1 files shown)


ronkorving:
I can confirm this solves the issue. And because bucket creation doesn't ignore auth errors, it's safe. But this looks really bad IMHO in terms of logic. Why the heck do we get an auth error when a bucket is in the process of being created? Why not just a "busy" error or whatever?

I think we need to take this to the Couchbase team.

stelcheck:
@ingenthr FYI (sorry for pinging you directly, wasn't sure who to refer to)

PR #172, opened 2017-11-20
Adds tests for archivist encoders

ronkorving:
This covers one commit of #171: https://github.com/mage/mage/pull/171/commits/f0332beec4a3d76b07736e805ec971d99e29f3d0

### DIFF
--- a/test/archivist/encodings.js
+++ b/test/archivist/encodings.js
@@ -1,0 +1,52 @@
+const assert = require('assert');
+
+describe('Encoding', function () {
+	const encoders = require('../../lib/archivist/encoders');
+
+	describe('can guess an encoding', function () {
+		it('"hello" -> utf8', function () {
+			assert.equal(encoders.guessEncoding('hello world'), 'utf8');
+		});
+
+		it('"foo" in a Buffer -> buffer', function () {
+			assert.equal(encoders.guessEncoding(new Buffer('foo')), 'buffer');
+		});
+
+		it('plain JS object -> live', function () {
+			assert.equal(encoders.guessEncoding({ a: 'b' }), 'live');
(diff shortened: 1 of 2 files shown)


test/archivist/encodings.js
```diff
@@ -0,0 +1,27 @@
+const assert = require('assert');
+
+describe('Encoding', function () {
+	const encoders = require('../../lib/archivist/encoders');
+
+	it('can guess an encoding', function () {
+		assert.equal(encoders.guessEncoding('hello world'), 'utf8');
```
stelcheck: Each assert being a single test here, it would be preferable to have a describe('can guess') with an it('') for each of those tests.

test/archivist/encodings.js
```diff
@@ -0,0 +1,27 @@
...
+
+	it('can do simple conversion between encodings', function () {
+		const mediaType = 'some/dummy';
+
+		function enc(from, to, data) {
+			const encoder = encoders.getEncoder(mediaType, from, to);
+			return encoder(data);
+		}
+
+		assert.deepStrictEqual(enc('utf8', 'buffer', 'こんにちは'), new Buffer('こんにちは'));
```
stelcheck: Same as above.

test/archivist/encodings.js
ronkorving: Then I think it would make more sense to add message arguments to the assertion, no?

stelcheck: The message wouldn't show on success cases.

stelcheck: e.g. by looking at a test run output, we cannot see what has been tested (which is instructive for any newcomer to the project)

PR #174, opened 2017-11-26
mage.quit, mage.exit

### LINKED ISSUE #140: Mage.quit() when cluster is true
This API is a bit inconsistent: from the master, it will quit the server, but from a worker it will trigger a recycle (since in that case, it is the worker that gets restarted).

stelcheck:
* Renamed `mage.quit` to `mage.exit`
  * Created a new `mage.quit` function that forwards the exit
    request to the master process.

Requires #151
Fixes #140

### DIFF
--- a/lib/mage/Mage.js
+++ b/lib/mage/Mage.js
@@ -26,7 +26,7 @@ function testEngineVersion(pkg, pkgPath, mage) {
 
 	if (!semver.satisfies(process.version, range)) {
 		console.error('Node.js version', process.version, 'does not satisfy range', range, 'as configured in', pkgPath);
-		mage.quit(-1);
+		mage.exit(-1);
 	}
 }
 
@@ -223,6 +223,7 @@ Mage.prototype.getTask = function () {
 
 Mage.prototype.setupCoreLibs = function () {
 	var core = this.core;
+	var that = this;
 
 	// Load the configuration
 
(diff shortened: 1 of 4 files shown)


lib/mage/Mage.js
```diff
@@ -262,19 +263,52 @@ Mage.prototype.setupCoreLibs = function () {
...
 
+/**
+ * Close down the MAGE server
+ *
+ * Workers will forward the request to their master if needed.
+ * This differs from `exit` since exit is equivalent to `process.exit`;
+ * see `exit` for more details.
+ *
+ * @param {number}  exitCode An exit code for node to return.
+ * @param {boolean} hard Wait for all services to closes before exiting
```
ronkorving: .. if false

stelcheck: You mean if true?

ronkorving:
If a module's setup causes an issue, does it not end up in an endless error-loop? Maybe that was already always the case though... Enlighten me please, oh wise one :)

lib/mage/Mage.js
ronkorving: `hard` being true sounds to me like it *wouldn't* wait, right?

stelcheck: .... ah. These were copy/pasted from the current doc, wasn't paying attention.

stelcheck:
I have module setup timeouts sometimes and have never seen this issue. What do you see that might change things in a way that it would be an issue now?

stelcheck:
Fixed jsdoc issues.

ronkorving:
Well that's the thing. I may have always been a problem.

What I saw was that module setup failure calls `mage.exit()`. So if a module always fails (due to some bug), the master process will never shut down in failure, I guess? It'll just keep rebooting workers until the universe ends or something?

I'm not saying this PR introduces an issue. I think it may have made an existing issue more obvious.

lib/mage/Readme.md
```diff
@@ -76,7 +76,11 @@ only argument.
 
 ### mage.quit(number exitCode)
 
-Shuts down MAGE and exits the process with the given exit code, or 0 if none has been passed.
+Shuts down MAGE and all its processes.
```
ronkorving: Might be nice to mention that the master process will exit with the given exit code.

stelcheck: The readme's should go at some point... so TBH I don't really want to spend too much time on them.

stelcheck:
About module setup: I don't believe this occured in the past, at least not on my projects (had cases of timeouts, for instance)

PR #176, opened 2017-11-26
Entity too large

stelcheck:
Allow to configure a max POST data size for the server.


stelcheck:
Just re-wrote the PR, please feel free to have a look

lib/httpServer/transports/http/index.js
```diff
@@ -615,6 +637,48 @@ function safeDecodePath(path) {
...
+		byteSize += chunk.length;
+
+		if (isRequestTooLarge()) {
+			req.emit('error', new RequestError('Entity too large', req));
+			req.removeAllListeners();
+		}
+	});
+
+	req.on = function (eventName, func) {
+		var eventName = arguments[0];
```
ronkorving: What's the purpose of this?

lib/httpServer/transports/http/index.js
```diff
@@ -615,6 +637,48 @@ function safeDecodePath(path) {
...
+	req.on('data', function (chunk) {
+		byteSize += chunk.length;
+
+		if (isRequestTooLarge()) {
+			req.emit('error', new RequestError('Entity too large', req));
+			req.removeAllListeners();
+		}
+	});
+
+	req.on = function (eventName, func) {
```
ronkorving: Why this hackery? Can't we just listen along on `req` and destroy the ongoing request when we reach the threshold?

ronkorving: I don't understand the purpose of the entire block that starts here.

lib/httpServer/transports/http/index.js
```diff
@@ -615,6 +637,48 @@ function safeDecodePath(path) {
...
+
+	function isRequestTooLarge() {
+		return byteSize > maxRequestSize;
+	}
+
+	req.on('data', function (chunk) {
+		byteSize += chunk.length;
+
+		if (isRequestTooLarge()) {
+			req.emit('error', new RequestError('Entity too large', req));
```
ronkorving: Does this also close the connection? Else we would still be receiving data, just not in userland.

lib/httpServer/transports/http/index.js
```diff
@@ -17,6 +19,21 @@ var handlebars = require('handlebars');
...
+/**
+ * The RequestError class can be used to emit or otherwise
+ * throw request-related errors.
+ */
+class RequestError extends Error {
+	constructor(message, req) {
+		super(message);
+
+		this.type = 'RequestError';
+		this.remoteAddress = req.headers['x-forwarded-for'] || req.socket.remoteAddress;
```
ronkorving: Maybe have a look at: https://www.npmjs.com/package/forwarded-for

scripts/templates/javascript-project/config/default.yaml
```diff
@@ -66,6 +66,10 @@ server:
     # the process to shut down.
     shutdownGracePeriod: 15
 
+    # Set the maximum HTTP POST (in KB) data request size for
+    # this server
+    maxRequestSize: 512
```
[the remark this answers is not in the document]
stelcheck: I've seen used elsewhere on this project, so sounds reasonable to use it here as well.

lib/httpServer/transports/http/index.js
stelcheck: Must be something I forgot to clean up. I'll remove it.

stelcheck: Line 666, how uncanny.

lib/httpServer/transports/http/index.js
stelcheck: I will be adding an extensive description in comment to explain this. Not ideal, but at least it should help the next person touching this code.

scripts/templates/javascript-project/config/default.yaml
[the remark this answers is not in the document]
stelcheck: Actually, we don't (not that I could find). Unless we'd use it everywhere, I am not sure I should be introducing this as part of this PR (could be a good idea as part of a PR dealing specifically with that though).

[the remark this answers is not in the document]
ronkorving: I think it could well be part of this PR, because you're introducing a new feature here. Introducing it later would mean changing the configuration schema, which you can avoid by doing it now.

lib/commandCenter/httpBatchHandler.js
```diff
@@ -243,6 +243,13 @@ function parseMultipart(req, boundary, cb) {
 	// pipe all incoming data straight into the multipart parser
 
 	req.pipe(parser);
+
+	req.on('error', function (error) {
+		delete cmdData;
+		delete files;
```
ronkorving: These two lines do nothing. `delete` is an operator to remove properties from objects.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

lib/commandCenter/httpBatchHandler.js
```diff
@@ -266,9 +273,13 @@ function parseSinglepart(req, cb) {
 
 	req.on('end', function () {
 		logger.verbose('Finished reading', req.method, 'request.');
-
 		cb(null, cmdData, null);
 	});
+
+	req.on('error', function (error) {
+		delete cmdData;
```
ronkorving: same

lib/httpServer/transports/http/index.js
```diff
@@ -615,6 +638,57 @@ function safeDecodePath(path) {
...
+		}
+
+		realOn('data', function (chunk) {
+			if (!isRequestTooLarge()) {
+				func(chunk);
+			}
+		});
+	};
+
+	return req;
```
ronkorving: Returning `req` makes it seem (from the call-site) that you may not get back what you put in. But you do. Perhaps not return anything?

lib/httpServer/transports/http/index.js
```diff
@@ -615,6 +638,57 @@ function safeDecodePath(path) {
...
+	function isRequestTooLarge() {
+		return byteSize > maxRequestSize;
+	}
+
+	req.on('data', function (chunk) {
+		byteSize += chunk.length;
+
+		if (isRequestTooLarge()) {
+			req.emit('error', new RequestError('Entity too large', req));
+			req.removeAllListeners();
```
ronkorving: `req.removeAllListeners('data');` just to be sure? This doesn't feel (but may well be) side-effect free.

lib/httpServer/transports/http/index.js
ronkorving: Have you tried using `req.destroy(new RequestError('Entity too large', req));`? I think it may be made for this.
I wonder if it would avoid the trickery that's happening below.

lib/httpServer/transports/http/index.js
stelcheck: We remove all event listeners because we also want to remove 'end' right now; we don't want 'end' to trigger if we have aborted.

Going through the list of events, I'd say removing everything *should* be safe enough - maybe even the safest thing to do. Can you see any events we'd actually want to fire on the stream post-error emission?

scripts/templates/javascript-project/config/default.yaml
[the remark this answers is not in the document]
stelcheck: Be that as it may, this is not how bytes are configured by standard in MAGE. I don't want to have people try to remember which config entry takes which kind of numbers.

lib/httpServer/transports/http/index.js
stelcheck: I have. It doesn't emit the error nor stop the 'data' emission.

lib/commandCenter/httpBatchHandler.js
stelcheck: So I guess the data will just naturally expire in this case? Are there any cleanup steps we should actively take?

lib/httpServer/transports/http/index.js
stelcheck: Should we change the behavior of this function in the future, we'd then need to change the call site as well. I prefer not relying explicitly on the existence of side-effects in general.

lib/httpServer/transports/http/index.js
ronkorving: No, it was more about future proofing and the fact that "get rid of all-the-things" is a bit of a broad sweeping approach. I doubt any Node documentation would ever argue that's a good idea to do on Node's internal APIs. To me, it's a bit like using underscored properties on Node internal APIs. Just because you can, doesn't mean you should (in that Node's internals may depend on you not "misbehaving").

lib/commandCenter/httpBatchHandler.js
ronkorving: I don't think you would have to do anything. The garbage collector should just clean it up. If it didn't, we would have a leak which would be a bigger problem regardless.

lib/httpServer/transports/http/index.js
ronkorving: Could that be worth filing an issue over?

The error should definitely be emitted according to the docs:

> Calls destroy() on the socket that received the IncomingMessage. If error is provided, an 'error' event is emitted and error is passed as an argument to any listeners on the event.

lib/httpServer/transports/http/index.js
stelcheck: Right, indeed Node's internal never use it (only case I found is https://github.com/nodejs/node/blob/c0d858f8bb8ba5212548da2fba6a7bc02db0462b/deps/npm/lib/search/all-package-metadata.js#L184 but its not  really relevant).

So I guess removing 'data' and 'end' would make more sense.

lib/commandCenter/httpBatchHandler.js
stelcheck: Right. Perhaps my question was more about whether we can clean up the data so that it doesn't have to wait for garbage collection, but I am guessing this isn't possible?

lib/httpServer/transports/http/index.js
stelcheck: Ah, apologies, I believe it does throw the error. The issue was that it also abruptly close the connection, not leaving the chance to external APIs to close/return what they wish. 

At any rate, either way does not stop the 'data' or 'end' data emission, however.

stelcheck:
All outstanding issues fixed, only one point remaining to be discussed.

scripts/templates/javascript-project/config/default.yaml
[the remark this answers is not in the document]
ronkorving: Do we have that many other cases where we specify size in bytes in config?

lib/httpServer/transports/http/index.js
ronkorving: I'm missing why abruptly closing the connection is a problem. Isn't that exactly what we would want?

That is why I asked in my initial comment:

> Does this also close the connection?

stelcheck: On error, upper levels can decide how to respond on their own terms.

Calling destroy returns *no response*, it just drops the connection.

ronkorving: Alright, I see what you're doing. I thought the premise from the beginning *was* to clear out the connection completely, with the assumption that data uploads that are wildly beyond an application's limits would be regarded abuse of some sort.

stelcheck: Since we are doing the filtering for *all* input on HTTP POST data, the idea is to let higher levels decide what error they wish to return. The idea behind this is that the communication protocol remains unknown to the HTTP server itself at this point (do we speak JSON? how should an error be returned?).

ronkorving: My idea of the right way to avert an attack or abuse of some sort would be to cut the connection and not care about communicating anything back to the client. We do not have a responsibility to clients with ill intentions. We can still log it server-side.

But if there are use-cases where you really want to be friendly to the client, then yeah, we would have to be a bit more friendly. In that case, HTTP status code 413 was made for this and would make for a fine default. I would let the HTTP status code take care of the communication, not the payload you send back (just don't send any).

stelcheck: Once the server sends a response, I understand that the upload will be interrupted anyway. Sending an error response that will be supported by the underlying communication protocol off the bat only adds minimal overhead, and makes it easier for protocol implementer to handle such cases without having to worry on top of that about HTTP headers.

scripts/templates/javascript-project/config/default.yaml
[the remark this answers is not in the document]
stelcheck: Not sure, I would have to scan the repository to find such cases - which is why I would argue it should be a task to be done separately.

lib/httpServer/transports/http/index.js
ronkorving: So then maybe just send back 413 and leave it there? Rather than giving a whole lot of freedom (and responsibility) to user land?

stelcheck: I more leaning towards providing a mean for higher levels to respond with an error within the protocol. This should make error management on the client end easier to do.

ronkorving: That adds more responsibility to server code though. In this PR, the command center will end up sending a `400 Bad Request` status back. That's less helpful than a `413` imho. So, I understand what you want to see, but then it would be nice if this PR demonstrated its proper use.

stelcheck: That structurally relies on requests sent over HTTP and demands that client implementations care about HTTP codes to do a good job. In the current case, I believe clients (and developers using them) won't have to change a single line of code and still receive a meaningful error that does not require them to understand HTTP. This means that they can code against it, which I don't believe they would be able to do should it be HTTP error code based.

Right now, we are sending "server" error though, so that would need to be changed.

ronkorving: I traced the logic back to it sending back a `400`? Did I miss something?

stelcheck: No, you are right indeed. So we'd need to fix that as well.

PR #179, opened 2017-11-27
devMode: removing options, removing session UC

stelcheck:
Removing loginAsActor, loginAnonymous and reassign from
session's module user commands.

Doing so made it easy to remove the following dev mode
configuration entries:

  * loginAs
  * customAccessLevel
  * adminEverywhere

Only one custom devMode configuration entry now remains
('archivistInspection').


ronkorving:
As far as I'm concerned, `archivistInspection` can just be `true` when development mode is active, period.

stelcheck:
I agree. I left it out of this PR because it can easily be removed at a later time (and I wasn't sure if removing was 110% the right move at the time).

PR #180, opened 2017-11-27
Pluggable vault backends

stelcheck:
Enable developers to import their own vault backend modules. This will
be the first step towards externalising many of the backend modules
currently in core.


Nokitoo:
If we make the developers import their own vault backend modules the same way we are doing with our modules I think it won't be convenient.
For each vault module they will need at least 3 files: index.js (class definition), Archive.js, and defaultTopicApi.js. It doesn't seem that defaultTopicApi and Archive are used elsewhere than in the vault itself.
I think they could just define a get/set/list method in the vault file, and choose how to separate their code.

PR #185, opened 2017-12-06
incorrect topic size check, check value size

stelcheck:
We were checking the size of a VaultValue, which
would cause sizeof to tight-loop because of circular
dependencies.

The checks for size were reversed.


ronkorving:
Travis doesn't seem to have much macOS processing available. It passes fine on Linux. I'm going to merge this.

PR #186, opened 2017-12-07
msgStream: Active session checking

### LINKED ISSUE #184: session: websocket for msgStream does not touch session
If I use long-polling, I will be refreshing my session on every call; however, this does not happen with websockets, despite the fact that we can still see the user connected.

stelcheck:
Idle msgStream connections can now request a check at a certain
interval. This is only used for session validation at the moment,
but it solves the following issues:

  1. Websocket
  2. Long polling timers bigger than the session ttl

Fixes #184


lib/modules/session/index.js
```diff
@@ -441,6 +441,18 @@ mage.core.cmd.registerMessageHook('mage.session', false, function (state, cfg, b
 
 
 mage.core.msgServer.msgStream.addAddressType('session', {
+	checkInterval: sessionTTL / 2 * 1000,
+	check: function (address, host, cb) {
+		try {
+			const fullSessionKey =	address.split('/')[1];
```
ronkorving: I think the only way parsing can fail is if `address` is not a string. In other words, maybe a simple `typeof` check would be more appropriate than this try/catch construction with an asynchronous function call inside.

lib/modules/session/index.js
```diff
@@ -441,6 +441,18 @@ mage.core.cmd.registerMessageHook('mage.session', false, function (state, cfg, b
 
 
 mage.core.msgServer.msgStream.addAddressType('session', {
+	checkInterval: sessionTTL / 2 * 1000,
+	check: function (address, host, cb) {
```
ronkorving: Would calling this `touch` or something not be more appropriate?
When I read `check` I thought it meant the same as `lookup`, but its purpose seems very different.

lib/modules/session/index.js
stelcheck: `try/catch` is what is currently being used elsewhere (see line 457 onward).

Regarding your second suggestion, will do.

lib/modules/session/index.js
stelcheck: It's a bit tricky since it not only extends but will fail to extend if the session has expired in the meantime. I'll rename it to extend anyway since I agree this makes better sense.

stelcheck:
Fixed code according to comments.

lib/modules/session/index.js
```diff
@@ -441,6 +442,18 @@ mage.core.cmd.registerMessageHook('mage.session', false, function (state, cfg, b
 
 
 mage.core.msgServer.msgStream.addAddressType('session', {
+	extendInterval: sessionTTL / 2 * 1000,
+	extend: function (address, host, cb) {
+		try {
+			const fullSessionKey =	parseAddress(address).fullSessionKey;
```
ronkorving: tab or full-width space?

lib/msgServer/msgStream/index.js
```diff
@@ -66,6 +69,36 @@ function MsgStream(cfg) {
...
+
+			this.intervals[type] = setInterval(function () {
+				async.each(that.getAddressesOfType(type), function (entry, cb) {
+					configuration.extend(entry.address, entry.host, cb);
+				}, function (error) {
+					if (error) {
+						logger.error(error);
+					}
+				});
+			}, extendInterval);
```
ronkorving: It would be nice and clean to unref the timer that is created here.

eg:

```js
this.intervals[type].unref();
```

ronkorving:
So, the sessions gets extended now periodically regardless of transport, but *also* on http longpolling heartbeat, which is a bit double. This may not be a huge deal, but still...

It also touches sessions after pulling their full object out of the database (even though that is not required, all we need to do is touch). Given that a lot of sessions will be dealt with in parallel and that in a distributed cluster, a lot of these intervals will fire at the same time (because they started at the same time), the load on the DB could be quite significant. Would love to hear your thoughts on that.

stelcheck:
As per description, it is by design: now, you can have a session TTL that is shorter than the long polling timeout (which also break right now).

As for touching the session... we could do a simple touch indeed.

stelcheck:
Addendum: I think we need to fetch the data to confirm that the session wasn't invalidated in the meantime.

ronkorving:
Any thoughts on the topic of the concentrated load that this will cause?

stelcheck:
Right, wasn't thinking about that, sorry. 

That's a problem indeed. We could make it into a per-connection basis, if we are OK with having more timers?

ronkorving:
Why not move the logic to the client SDK instead? For longpolling, no extra load is needed (maybe the interval can be smarter, but I personally think it's fine like this... I don't expect sessions shorter than 30 sec). For WebSockets, the clients could pro-actively send a heartbeat to the server. That way all touch-operations would *not* be synced and load would be well distributed.

stelcheck:
This would not solve the "session shorter than long-polling timeout" issue (I actually ran into this).

ronkorving:
> maybe the interval can be smarter, but I personally think it's fine like this... I don't expect sessions shorter than 30 sec

stelcheck:
A session expiration of 30 sec would still break with the default long polling timeout (60 sec IIRC). At any rate, I had a case where I was testing session expiration by setting session timeout to 5 seconds, so while these cases aren't likely in production, they are still somewhat likely as part of the development lifecycle (at least that's how I ended up running into the issue).

In regards to having the client send keepalive, my concern is with adding yet another requirement for client SDK developers to be aware of. It's a possible fix, but I'd prefer finding a way to solve this without the clients having to necessarily do anything about it, ideally.

ronkorving:
I can appreciate the goal of trying to solve this server-side and keep the clients as dumb as we can get away with. But from a technical perspective, I do think solving this client-side will end up creating a superior user experience while keeping server load in check. Maybe there are some things I'm missing of course, but right now... I'm out of ideas beyond what I suggested.

stelcheck:
We'd need to add some meta-messages to the msgStream protocol - that is, messages which have the sole purpose of pinging the session while returning nothing. 

We wouldn't be able to do this on long-polling, which means we'd need to compute the long polling timeout to be smaller than the session timeout (and potentially add some log entry to warn the developer of the adjustment made). On WebSocket, we would need to implement this through ping or metadata messages.

The two weaknesses I see are:

  1. Each transport will have a different logic and will need to implement keepalive, which means
  2. On the client, each client implementation will be different per transport

Arguably, this would be a more complete - or correct - implementation, since it would also take into account connectivity. But it would be way more involved. 

Right now, the implementation proposed in this PR basically would work for any transport implementation we might choose to add in the future while making it relatively easy to augment to do what you suggest in the future should it be desirable at some point.

ronkorving:
I agree those are the changes that would need to happen, and they're not ideal at all from a maintenance point of view.

But my concern regarding load stands... You made one suggestion: a single timer per connection. I think that would be manageable and maintainable. Let's go with that then 👍

PR #187, opened 2017-12-07
Removing Plato

### LINKED ISSUE #4: What to do about Plato?
Right now we use the `yaml2json` dependency for Plato, which seems overkill. Plato should be able to pass a yaml config file to ESLint. If it can't, let's fix Plato.

We also have `yamljs` lingering around, and it seems we should not need it.

If we can get rid of them, that would be nice. But these are dev dependencies only and don't have any negative impact (beyond dev installations being a tiny bit slower), so it's not a massive priority.

stelcheck:
Fixes #4


package.json
```diff
@@ -27,16 +27,12 @@
     "preversion": "npm run docs:build && git add ./docs",
     "report:coveralls": "istanbul cover _mocha --root lib --include-all-sources --report lcovonly --dir coveralls -- ./test/index.js && cat ./coveralls/lcov.info | coveralls && rm -rf ./coveralls",
     "report:coverage": "istanbul cover _mocha --root lib --include-all-sources --report html --dir coverage-report -- ./test/index.js",
-    "report:complexity:create-config": "yaml2json .eslintrc.yml > .eslintrc",
```
ronkorving: I think this means we can get rid of the `yamljs` dev dependency.

stelcheck: Indeed, just removed it.

PR #189, opened 2017-12-08
serviceDiscovery: do not mess with the IP order

stelcheck:
Changing the IP order breaks how clustering works on Docker Swarm.

This is due to the fact that interfaces have *two* IP addresses:

  - One being the IP being used for communication on the overlay network
  - One being used locally for local routing on the local swarm node

Connecting to that second interface results in MMRP and
other services relying on service discovery to essentially connect
to themselves.

I also took the opportunity to uniformize how Zookeeper and Consul
work (or at least how they select the IP to use, and allowing
similar configuration).

### DIFF
--- a/lib/serviceDiscovery/engines/consul/index.js
+++ b/lib/serviceDiscovery/engines/consul/index.js
@@ -5,9 +5,10 @@ const os = require('os');
 const requirePeer = require('codependency').get('mage');
 const consul = requirePeer('consul');
 
-const mage = require('../../../mage');
+const mage = require('lib/mage');
 const Service = require('../../service').Service;
 const ServiceNode = require('../../node').ServiceNode;
+const helpers = require('../../helpers');
 
 const logger = mage.core.logger.context('serviceDiscovery', 'consul');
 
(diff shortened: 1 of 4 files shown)


lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -5,9 +5,10 @@ const os = require('os');
 const requirePeer = require('codependency').get('mage');
 const consul = requirePeer('consul');
 
-const mage = require('../../../mage');
+const mage = require('lib/mage');
 const Service = require('../../service').Service;
 const ServiceNode = require('../../node').ServiceNode;
+var helpers = require('../../helpers');
```
ronkorving: That `var` feels out of place :)

ronkorving:
It's for this kind of thing that I wrote https://github.com/Wizcorp/network-interfaces :)

lib/serviceDiscovery/helpers.js
```diff
@@ -8,35 +8,95 @@ const interfaces = os.networkInterfaces();
...
+		return getAnnouncedIpsForInterface(ifaceName);
 	}
 
-	return announceIps.sort();
+	const announced = ifaceNames.reduce((announced, name) => pushIps(interfaces[name], announced, false), []);
+	assertNotEmpty(announced, 'Could not find any valid IP addresses');
+
+	// Note: the list of IP **MUST** stay in the order received by
+	// `os.networkInterfaces()`; otherwise, running MAGE on a Docker Swarm
+	// will break in some cases.
```
ronkorving: And yet the order cannot be guaranteed by virtue of how JavaScript works (ie: we're lucky V8 keeps it in a certain order, but even Node cannot guarantee that will remain the case).

Could we not filter with a configurable mask for example?

stelcheck: The order of the interfaces cannot, but the array of IP for each interface can.

stelcheck:
We cannot use this library because it uses default values for parameters, which is not valid on Node 4. Once we remove support for Node 4 (any time now I would expect), we'll be able to start using it I think.

lib/serviceDiscovery/helpers.js
ronkorving: Ah, and those were the problem? Got it 👍

PR #192, opened 2017-12-11
ArchivistModuleError, module refactoring

stelcheck:
Refactored the code to use promises, reduce copy/pastes, and
throw ArchivistModuleError which contain more details for
logging purposes.


stelcheck:
istanbul needed to be upgraded to alpha so that fat arrow functions would be correctly procesed.

lib/modules/archivist/ArchivistModuleError.js
```diff
@@ -0,0 +1,19 @@
...
+	constructor(message, state, details) {
+		super({
+			message: message,
+			details: Object.assign({
+				actorId: state.actorId,
+				acl: state.acl
+			}, details)
+		});
+
+		this.type = ArchivistModuleError;
```
ronkorving: What's the purpose of this? `error.constructor` === `error.type`

lib/modules/archivist/usercommands/rawApplyDiff.js
```diff
@@ -1,45 +1,19 @@
...
-				'Actor ' + state.actorId + ' not allowed to set diff value.',
-				cb
-			);
-		}
-
-		try {
-			value.applyDiff(diff);
-		} catch (e) {
-			return state.error(null, e, cb);
+			throw new ArchivistModuleError('Actor not allowed to set diff on this topic', state, topic, index);
```
ronkorving: Wouldn't you want to know *who* did this?

stelcheck: The state contains the actorId and that actorId will be logged.

lib/modules/archivist/ArchivistModuleError.js
stelcheck: Ah, I think that should be a string. Ill fix it

ronkorving: Then I would suggest having `MageError` automatically assign `this.type = this.constructor.name;`, so you never have to type that line of code again.

stelcheck:
Everything should be fixed.

lib/modules/archivist/index.js
```diff
@@ -1,2 +1,48 @@
...
+	const topicApi = state.archivist.getTopicApi(topic, clientVault.name);
+	if (!topicApi) {
+		throw new ArchivistModuleError('Topic does not exist or has no client configuration', state, {
+			topic: topic,
+			index: index
+		});
+	}
+
+	return topicApi;
+};
```
ronkorving: For getTopicApi and getClientVault.. why not just have the existing functions throw? I can imagine you're trying to stay BC, but it seems a bit overkill and inconsistent to have these in 2 places.

lib/modules/archivist/usercommands/rawSet.js
```diff
@@ -1,50 +1,23 @@
...
-	var topicApi = state.archivist.getTopicApi(topic, clientVault.name);
-	if (!topicApi) {
-		return state.error(
-			null, 'Unable to load topic API on client vault for topic: ' + topic, cb
-		);
-	}
+exports.acl = ['*'];
 
-	var options = {
+exports.execute = async function (state, topic, index, data, mediaType, encoding, expirationTime) {
```
ronkorving: Will `async` not break on Node 6?

stelcheck: Yes, it would. It takes 7 with a flags and 7.6 and up without flag. 

I'll fix that.

lib/modules/archivist/index.js
stelcheck: The goal was to make the errors being logged more consistent (e.g. if you hit archivist client API you always get more or less the same kind of errors). So it's a bit redundant, but it does currently serve a purpose.

lib/modules/archivist/usercommands/rawSet.js
stelcheck: Going back to Promises-based doesn't really help us in terms of readability here. I'd rather wait for this to get merged once we remove support for Node <8.

PR #195, opened 2017-12-16
Make user command cache optional

stelcheck:
User commands may now be configured to
not cache requests.

### DIFF
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -244,7 +244,8 @@ CommandCenter.prototype.setupUserCommand = function (modName, cmdName, cmd) {
 		// for external wrapper libraries which may be using the rest
 		// operator
 		params: cmd.params || params,
-		isAsync: isAsync
+		isAsync: isAsync,
+		cache: cmd.cache !== false
 	};
 };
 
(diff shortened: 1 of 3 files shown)


stelcheck:
I have tested that this doesn't break the normal flow, and that caching gets triggered as expected, but I have **not** tested retries yet. Would appreciate some help with that.

lib/commandCenter/commandCenter.js
```diff
@@ -788,9 +789,18 @@ CommandCenter.prototype.executeBatch = function (acceptedEncodings, batch, query
...
+			function runCommand(output, cache, cmdIndex) {
 				var cmd = batch.commands[cmdIndex];
 				if (cmd) {
+					if (cacheContent.length !== 0 && isCachedCommand(cmd)) {
+						logger.warning
+							.data({ cmd: cmd.cmdName })
+							.log('Re-sending cached command response');
+
+						output += (output ? ',' : '[') + cacheContent.shift();
+						return setImmediate(runCommand, output, cache, cmdIndex + 1);
```
ronkorving: The signature here and on line 823 seem to disagree with each other.

stelcheck: Fixed the signature.

PR #198, opened 2017-12-22
Added change password feature to MAGE

Brunni132:
```
auth.register(state, username, password, options, function (error, userId) {
	auth.changePassword(state, username, newPassword, function (error) {
		auth.login(state, username, newPassword, function (error, session) {
			// Profit
		});
	});
});
```

### DIFF
--- a/docs-sources/includes/_Actors.md
+++ b/docs-sources/includes/_Actors.md
@@ -125,3 +125,17 @@ EOF
 
 If authentication fails, you will receive `[["invalidUsernameOrPassword",null]]`;
 otherwise, you should get back an event object containing your player's session information.
+
+## Changing the password
+
+You can change the password for a given user after registration. It works the same way as registration, except
+that the user ID needs to exist or you will get a `UserNotFoundError`. Note that this does not invalidate current
+sessions for this account.
+
+>  lib/modules/players/index.js
+
+```javascript
+exports.changePassword = function (state, username, newPassword, callback) {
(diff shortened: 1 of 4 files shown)


ronkorving:
Looks good! Could you also add this new method to the ReadMe in this folder?

stelcheck:
@ronkorving shouldn't it go in the user doc instead perhaps? https://mage.github.io/mage/#auth-module

Brunni132:
Yup thanks guys! I'll do all that.
Btw was wondering, is there any limitation to the passwords currently in MAGE? Or in Kakegurui or any depending project? I haven't tested but looking at the code it would look like '' is a perfectly valid password??

stelcheck:
I don't think the auth module sets any limitations, if thats what you mean

Brunni132:
OK thanks!

Not exactly sure what you mean about the ReadMe, and what you would like to see and where. I added the documentation in the mage.ts, so that it would appear online and can be used from typescript. Please tell me if that's not enough.

stelcheck:
I would add usage documentation to https://mage.github.io/mage/#auth-module as well for completeness. This documentation is generated from https://github.com/mage/mage/blob/master/docs-sources/includes/_Actors.md#auth-module

Brunni132:
OK. I didn't bother because I didn't really grasp how doc is supposed to be organized.

In any case this section (mage-auth) is far from being comprehensive, isn't it? It only shallowly covers logging in (not register or what), and I fail to see where that process is described. Shouldn't that go along?

PR #199, opened 2017-12-22
Removed Readme files

stelcheck:
The ones which contain about the internal architecture of MAGE or that
are in subsystem that we plan to externalized were left in place.


ronkorving:
Is all of the information that is removed here covered by https://mage.github.io/mage/ ?

stelcheck:
Either that or it is no longer relevant or accurate.

PR #200, opened 2018-01-05
Remove Node 4 support

ronkorving:
When using versions of NPM too old to understand `package-lock.json` (or stick to its rules, I'm not entirely sure), they will end up installing dependencies of dependencies of versions that are not consistent with our package-lock file.

That can in turn mean that node modules with known vulnerabilities get installed, and we would like to avoid that. For that reason I had to remove Node 4 from our test matrix. Not all our current dependency versions are ready for Node 9 yet, so I have not been able to add Node 9 to the matrix.

That means our test matrix will become:

- Node 6
- Node 7
- Node 8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ before_install:
   - if [[ "$TRAVIS_NODE_VERSION" == "4" ]]; then npm install -g npm@3; fi
 
 script:
+  - travis_retry npm run test:lint
+  - travis_retry npm run test:security
   - travis_retry npm run report:coveralls
 
 env:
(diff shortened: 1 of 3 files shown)


stelcheck:
Why do we need to remove Node 4 as part of this PR?

stelcheck:
These 2 commits are independent, please do not squash.

There are more commits now, should they too  remain unsquashed?

ronkorving:
@stelcheck Let's discuss offline first

stelcheck:
Could you at least update the PR description so that it describes the intent?

PR #201, opened 2018-01-09
Updates a number of subdependencies

ronkorving:
This is a followup on #200, which did not update all the sub-dependencies we wanted to see bumped. I hope this does the trick.


ronkorving:
Oops, forgot, nice catch 👍

PR #208, opened 2018-02-19
Documentation improvement on supported Node versions.

### LINKED ISSUE #207: Recommended MAGE version
https://mage.github.io/mage/?shell#requirements

We need to change the content of the notice recommending Node 4.

@ronkorving I believe you removed Node 4 in a previous PR, could you take care of updating the documentation as well?

ronkorving:
This advocates the use of the most current LTS version and links to the LTS schedule.
Since we no longer support Node 4, this PR also removes it from the "engines" field in package.json.

Sidenote: it seems some documentation was altered, but never rebuilt. Since a rebuild of docs took place, they are a part of this PR.

Closes #207

### DIFF
--- a/docs-sources/includes/_Requirements.md
+++ b/docs-sources/includes/_Requirements.md
@@ -3,18 +3,22 @@
 ## Node.js
 
 <aside class="notice">
-We recommend using version 4 since it is an LTS version
+MAGE always supports the "Active LTS" (Long Term Support) versions of Node.js.
+We recommend using the most recent Active LTS version, but we always try to support the Current Release as well.
+
+You can see what the LTS and Current versions are in the Node.js
+<a href="https://github.com/nodejs/Release#release-schedule">release schedule</a>.
 </aside>
 
 ```shell
-nvm install 4
-nvm use 4
+nvm install 8
+nvm use 8
 ```
 
 ```powershell
 # You will need to provide the specific version to install and use
-Install-NodeVersion v4.8.2
-Set-NodeVersion v4.8.2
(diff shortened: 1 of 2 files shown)


docs-sources/includes/_Requirements.md
```diff
@@ -2,19 +2,20 @@
 
 ## Node.js
 
-<aside class="notice">
```
ronkorving: I had to remove the aside, since it seems that it doesn't support links in its content.

stelcheck: Surely `<a href="...">...</a>` should work?

stelcheck: Also, where would you need a link?

stelcheck:
The documentation is supposed to only be rebuilt on releases, no?

ronkorving:
@stelcheck Ah, fair enough. I rebuilt it to see what the resulting HTML would be, and committed it in. I can undo that if it helps?

docs-sources/includes/_Requirements.md
ronkorving: Link to the LTS schedule

stelcheck:
Probably better off to make sure the documentation matches what has been released.

docs-sources/includes/_Requirements.md
ronkorving: Do you want me to put the notice back? With the added information, I'm not sure if you still prefer that there or not.

stelcheck: I think that would be preferable.

PR #211, opened 2018-03-09
This allows Node 9 to be used

ronkorving:
It also removes Node 7 from the test-matrix (even though technically we can support it), so we can focus on the important versions. 7 has been superceded by 8 and is not maintained (ie: use at your own risk).

This bump was previously prevented by sqlite3 not being compatible with Node 9. It is now, however, so this PR bumps the sqlite3 version accordingly.

Closes #204

### DIFF
--- a/.retireignore
+++ b/.retireignore
@@ -1,6 +1,3 @@
-@brace-expansion
-
-# typedoc
 node_modules/typedoc-default-themes/bin/default/assets/js/main.js
 
 coverage-report/
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ dist: trusty
 language: node_js
 node_js:
   - 6
-  - 7
   - 8
+  - 9
 
 #
 # The last test is required to avoid issues with peer dependencies upon
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 environment:
   matrix:
+    - nodejs_version: '9'
     - nodejs_version: '8'
-    - nodejs_version: '7'
     - nodejs_version: '6'
 
 install:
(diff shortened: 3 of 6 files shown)


ronkorving:
This is currently the cause of the vulnerability blocker: https://github.com/mapbox/node-pre-gyp/issues/346

ronkorving:
A new version of node-pre-gyp has been released, but I don't think node-sqlite3 uses it yet.

stelcheck:
At first glance it appears like we should be able to safely ignore those. Can you confirm?

stelcheck:
Alternatively we can wait until #180 is completed and merged.

ronkorving:
Whichever lands first I guess :)

While safe to ignore, it marks all builds as broken from hereon. I don't think that's really desirable.

stelcheck:
I believe I mentioned this to you last time we had this issue, but you can ignore those errors by adding entries to the .retireignore file.

ronkorving:
Ah yes indeed, but what will trigger us to pull it out again? Or should we make a rule of adding all dev dependencies to .retireignore?

stelcheck:
We are probably better off adding problematic dependencies that can safely be ignored on a case-per-case basis. We should aim at ignoring as little as possible.

ronkorving:
Progress being made on sqlite3: https://github.com/mapbox/node-sqlite3/pull/952

ronkorving:
I've suppressed "low" level security warnings. Maybe that will do for now.

I couldn't get exceptions to work:

- I didn't want to outran ignore an entire package like lodash, but retireignore doesn't support version specificity.
- Ignoring a package does not retire ignore its children (so can't suppress `sqlite3` in order for some child down the tree not to cry).
- I pointed at the package of the insecure version in its specific location, but it was not being ignored. Couldn't get this to work.

ronkorving:
https://github.com/mapbox/node-sqlite3/blob/master/CHANGELOG.md#400 :)
Not yet on NPM though.

stelcheck:
Great, please make a PR once it is (supposing that should be momentarily)

PR #212, opened 2018-03-14
Remove unused function msgServer.getMsgStreamUrl()

ronkorving:
This API was used in the past by build systems, but has long been replaced by `getPublicConfig`.


stelcheck:
Not sure why but CI fails on security checks.

ronkorving:
Most (all?) dev dependency related:

- retire (the irony :))
- broken-link-checker-local
- coveralls

At least these seem to have dependencies that are reported.

PR #215, opened 2018-04-04
Add index to archivist.get

Nokitoo:
Add index key into archivist.get results


stelcheck:
As per our discussion, I will be closing this PR. The original goal is to make sure that `archivist.scan` calls return the topic's associated index. This would achieve that, but would also introduce breaking changes on a core API that is heavily used.

Instead, we should make a change on `archivist.scan` directly, which should essentially become a separate PR.

Let me know if I can help you with anything else!

PR #216, opened 2018-04-05
Add index to archivist.sca

Nokitoo:
Add index to archivist.scan results.

It changes the actual results:
`[ { name: "foo" }, { name: "bar" } ]`
to:
`[ [{ userId: 1 }, { name: "foo" }], [{ userId: 2 }, { name: "bar" }] ]`

### DIFF
--- a/lib/archivist/index.js
+++ b/lib/archivist/index.js
@@ -694,7 +694,15 @@ Archivist.prototype.scan = function (topic, partialIndex, options, cb) {
 		}
 
 		const queries = indexes.map((index) => ({ topic, index }));
-		this.mget(queries, options, cb);
+		this.mget(queries, options, (error, data) => {
+			// Add index to data
+			// Eg. [ [{ userId: 1 }, { name: "foo" }], [{ userId: 2 }, { name: "bar" }] ]
+			const indexedData = data.map((obj, idx) => {
+				return [queries[idx].index, obj];
+			});
+
+			cb(indexedData);
+		});
 	});
 };
 


stelcheck:
LGTM. Could you add a unit test to describe this new behaviour?

@christopherobin @AlmirKadric is this new behavior good with you?

AlmirKadric:
@Nokitoo Looks fine to me, but any reason to use an array as opposed to a more verbose object?
e.g. `[{ value: {}, index: {} }, { value: {}, index: {} }]`

Nokitoo:
You are right @AlmirKadric, a more verbose seems better to me. There was no special reason to choose an array.
What do you think @stelcheck ?

stelcheck:
@AlmirKadric because an array will fit more naturally with existing standard JS APIs when destructuring.

Example:

> Object iterations

```javascript
for (const [key, value] of Object.entries({ a: 1 }))
```

> Map Iterations

```javascript
const map = new Map()
map.set('a', 1)
for (const [key, value] of map.entries())
```
Since I think the logical use of scan data is either to return directly or to iterate upon, returing data that behaves more similarily to other APIs seems to make sense.

lib/archivist/index.js
```diff
@@ -694,7 +694,15 @@ Archivist.prototype.scan = function (topic, partialIndex, options, cb) {
 		}
 
 		const queries = indexes.map((index) => ({ topic, index }));
-		this.mget(queries, options, cb);
+		this.mget(queries, options, (error, data) => {
+			// Add index to data
+			// Eg. [ [{ userId: 1 }, { name: "foo" }], [{ userId: 2 }, { name: "bar" }] ]
+			const indexedData = data.map((obj, idx) => {
+				return [queries[idx].index, obj];
```
stelcheck: You could map against the indexes array instead, would make the code shorter I think.

Nokitoo: Right

Nokitoo: Right

stelcheck:
I understood your point. Mine was that destructuring on an object would look less like other existing standard APIs than using an array. See the code samples provided earlier.

AlmirKadric:
@stelcheck then I missing something from your example since the sample you provided does not seem to mimick what the data will be returning in the first place. Mind providing a sample which is similar to what @Nokitoo and I were discussing? e.g. why are you iterating an object? `Object.entries({ a: 1 })` / `map.entries()` the collection is always an array.

Nokitoo:
@AlmirKadric What he meant is that `[ [], [] ]` follow the standards of Javascript APIs like Object.entries() and Map.entries().
They both use arrays of length 2 for index and value.

AlmirKadric:
ah right, thanks for the clarification @Nokitoo . Does make a lot more sense now, I was wondering why the map and object entries came into play. I don't completely agree with the choice, but as there is a legitimate justification and my disagreement is more about personal taste LGTM 👍

Issue #206, opened 2018-01-31
Vaults read/write order


stelcheck:
https://mage.github.io/mage/#archivist does not document that vault backends need to be added to list, read and write order.

Nokitoo:
fixes #217

PR #217, opened 2018-04-05
Add warning for vaults list, read and write order

Nokitoo:
Add warning in doc to specify vaults need to be added in list, read and write order, as discussed in #206.

### DIFF
--- a/docs-sources/includes/_Archivist.md
+++ b/docs-sources/includes/_Archivist.md
@@ -9,20 +9,24 @@ uses Archivist behind the scenes to store credentials for newly created users.
 
 ## Vaults
 
+
+<aside class="warning">
+Never configure multiple vaults to connect to the same vault backend storage!
+This would break how [migration scripts](#migrations) work.
+</aside>
+
+<aside class="warning">
+Not all vaults support every operations! Below you will find
+a short configuration description for each available vault backends
+alongside a list of supported operations for that backend.
+</aside>
+
+### List, read and write order
+
 > ./config/default.yaml
 
 ```yaml
 archivist:
-    vaults:
-        userVault:
-            type: file
-            config:
-                path: ./filevault/userVault
-        itemVault:
-            type: file
-            config:
-                path: ./filevault/itemVault
-
     # When doing "list" operations, will attempt each mentioned vault until successful
     listOrder:
         - userVault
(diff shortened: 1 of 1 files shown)


Nokitoo:
Fix #206

stelcheck:
I think it would be nice to create a separate subchapter to explain read/write order; if we only put a warning, I fear end users won't fully understand what do do, and what purpose the system takes.

Nokitoo:
@stelcheck does this structure seems relevant to you ?
=> Vaults
     => List, read and write order
     => Vaults types

stelcheck:
Yes, I think that would work.

docs-sources/includes/_Archivist.md
```diff
@@ -22,6 +22,21 @@ archivist:
             type: file
             config:
                 path: ./filevault/itemVault
+
+    # When doing "list" operations, will attempt each mentioned vault until successful
```
stelcheck: I'd separate that portion into its own code sample block.

docs-sources/includes/_Archivist.md
```diff
@@ -35,6 +50,14 @@ a short configuration description for each available vault backends
 alongside a list of supported operations for that backend.
 </aside>
 
+
+### List, read and write order
+
+In `./config/default.yaml`, listOrder, readOrder, and writeOrder properties **have to** be defined to specify the order when reading or writting data.<br>
```
stelcheck: This can be specified and customized in any configuration file - it is not limited to default.

docs-sources/includes/_Archivist.md
```diff
@@ -35,6 +50,14 @@ a short configuration description for each available vault backends
...
+For read operations (listOrder, readOrder), mage will attempt the mentioned vaults until one is successful. However, for write operations (writeOrder), mage will write to each mentioned vault, in the given order.
```
stelcheck: The system seems to be a bit more elaborate: I *think* `state.archivist.distribute` will run multiple write operations in the configured write order.

https://github.com/mage/mage/blob/master/lib/archivist/index.js#L849

This means two things:

  1. Like you describe, topics configured to write to more than one backend will write to those backends in that order
  2. If you write multiple topics within a given state transaction, then they will be written to their respectively configured vault backend in the ordered provided by configuration

Nokitoo:
@stelcheck I updated the documentation.
I added multiple `<br>` otherwise the code block would merge with the block code above.

Nokitoo:
Can close #206

Nokitoo:
Closes #206

Issue #197, opened 2017-12-20
Confusing auth module in documentation


Nokitoo:
Hi,

I find the auth module is confusing in the documentation.
It's not said that it's not enabled by default, but it's used in some examples, before explaining how to enable auth module.

It's the case in the examples of the [modules-methods](https://mage.github.io/mage/#module-methods) part, [using-async-await-node-7-6](https://mage.github.io/mage/#using-async-await-node-7-6) part, and the [login](https://mage.github.io/mage/#login) part.

stelcheck:
Indeed. We should have fixed the documentation at the time we removed the auth built-in module from the list of default module.

I'll add a section on built-in modules and explain how to set up the auth module shortly.

Nokitoo:
@stelcheck What do you think about this structure ?

- Modules
	- built-in modules
		- Explain what built-in modules are
		- Specify already activated modules on fresh mage install
		- Explain auth module is not activated by default and explain how to setup
	- File structure
	- Module setup and teardown


- Actors and session
	- Session module
	- Auth module  <= Remove setup explanation and add link to redirect on built-in modules
	- Logging in

stelcheck:
Sounds good to me.

PR #218, opened 2018-04-06
Add built-in modules doc

Nokitoo:
Add documentation for built-in modules as referred in #197.


docs-sources/includes/_Modules.md
```diff
@@ -1,5 +1,70 @@
...
+  }
+};
+```
+
+> config/default.yaml
+
+```yaml
+module:
+    auth:
+        // this should point to the topic you created
```
stelcheck: This is not valid in YAML, use # not //

docs-sources/includes/_Modules.md
```diff
@@ -1,5 +1,70 @@
...
+        topic: auth
+        // configure how user passwords are stored, the values below are the
+        // recommended default, see the module's README.md for more details
```
stelcheck: Could we point to the documentation instead?

PR #224, opened 2018-04-18
Couchbase error helper: error may be null

stelcheck:
The helper must check whether the received value is an object
before attempting to read its attribute.


stelcheck:
@Nokitoo can you double-check that this does indeed prevent the issue mentioned above?

Nokitoo:
I prefer returning explicit values when the function is not void, but otherwise LGTM.
(I don't mind only checking for null, but didn't want you to check if it's an object ?)

stelcheck:
I think null check only should be fine.

Nokitoo:
LGTM then

PR #225, opened 2018-04-18
Add logger default config

### LINKED ISSUE #205: Errors in config file can cause the logger to die and report no errors, stopping the process with no output
For example, if you use this as your config, the empty 'server:' 
and then run 'npm run archivist:create'
mage will close with no errors
'''
logging:
    # Logging in the browser
    html5:
        # Log to the browser console
        console:
            channels:
                - ">=error"

        # Send log messages back to the server (usually errors and up)
        server:
            channels:
                - ">=error"

    server:
        # You should add logging services such as Graylog or Syslog here
'''

Nokitoo:
Add default config for the logger is none is specified, the same way default configs are handled with other Mage modules.

When specified no default config (specially logging.server undefined), it made the engine crash and no log was print.

- Overwrite logger config if not defined
- Changed Matryoshka.merge to allow null objects to be overwritten
  - Must specify false to completly disable a config
- Add unit tests to test this new behaviour
- Add default production config for logger

Fixes #205


scripts/templates/typescript-project/config/production.yaml
```diff
@@ -28,7 +28,11 @@ logging:
 
     server:
         # You should add logging services such as Graylog or Syslog here
-
+        terminal:
+            channels: ">=error"
```
stelcheck: Should be notice or higher

scripts/templates/javascript-project/config/production.yaml
```diff
@@ -28,6 +28,11 @@ logging:
 
     server:
         # You should add logging services such as Graylog or Syslog here
+        terminal:
+            channels: ">=error"
```
stelcheck: Same here

lib/config/Matryoshka/index.js
```diff
@@ -286,6 +286,33 @@ function merge(a, b) {
...
+	// Ex:
+	//		logging:
+	//			server:
+	//			||
+	//			\/
+	//		logging: {
+	//			server: null
+	//		}
+	if (a.value === null) {
+		return b.copy();
```
stelcheck: Any chances both a and b could be null?

PR #228, opened 2018-04-20
Separate vault connection from setup

### LINKED ISSUE #227: archivist:create is now broken for couchbase
Using couchbase 4.

```yaml
archivist:
  vaults:
    volatileVault:
      type: couchbase
      config:
        create:
          adminUsername: admin
          adminPassword: hello
          bucketType: couchbase
          ramQuotaMB: 100
          replicaNumber: 0
          bucketType: memcached
        options:
          scheme: couchbase
          hosts:
            - localhost:8091
          bucket: volatile
```

When I call `npm run archivist:create`, I get the following error:

```plaintext
Failed to load MAGE application: CouchbaseError: Authentication failed. You may have provided an invalid username/password combination
```

Nokitoo:
When starting the server, vaults initialization needs now 2 steps:
- call .setup on all vaults
- call .open on all vaults

When calling archivist:create, it called all vaults .setup, making CouchBase vault trying to open an inexistent bucket.

Fixes #227


lib/archivist/configuration.js
```diff
@@ -695,3 +695,20 @@ exports.setup = function (_logger, cb) {
...
+		Object.keys(persistentVaults),
+		function (vaultName, callback) {
+			const vault = persistentVaults[vaultName];
+
+			if (!vault.open) {
+				logger.error(`open method missing for vault ${vault.constructor.name}`);
+				return callback();
+			}
+
+			vault.open(callback);
```
stelcheck: `connectVaults` and `open` ends up using two words for the same action; we should be able to connect/disconnect vaults one at a time too, so we should make sure the API naming convention is consistent

lib/archivist/index.js
```diff
@@ -25,6 +25,10 @@ exports.setup = function (cb) {
 	configuration.setup(logger, cb);
 };
 
+exports.connectVaults = function (cb) {
```
stelcheck: We should also have disconnectVaults

Nokitoo: As we have closeVaults, I will name it openVaults

lib/archivist/configuration.js
Nokitoo: Same as lib/archivist/index.js comment

PR #232, opened 2018-04-23
Add default ttl for topics

### LINKED ISSUE #163: A default TTL (expiration time) for topics
We could add a default TTL (number and maybe even function) to topic/vault in lib/archivist/index.js, I think that would be very useful for cache lifetime management (configure it once, not for every `set()` call) and it can even replace the responseCache configuration completely.

Thoughts?

Nokitoo:
In lib/archivist/index.js, the user can specify a ttl string (See timeCodeToSec in lib/helpers/helpers.js for format) to expire the topic at a given time.

Resolves #163

### DIFF
--- a/lib/archivist/index.js
+++ b/lib/archivist/index.js
@@ -194,6 +194,15 @@ function attemptReadFromVault(state, vault, value, cb) {
 	});
 }
 
+function getTopicExpirationTime(topic) {
+	const config = configuration.getTopicConfig(topic);
+	if (config.ttl) {
+		return Date.now() / 1000 + config.ttl;
+	}
+
+	return null;
+}
+
 
 // Archivist instances manage access to vaults
 // -------------------------------------------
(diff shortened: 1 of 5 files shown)


Nokitoo:
As stated by ron in the issue #163, we could also add a function for the ttl. However, I don't see in which specific use case we would need it. Do you have an idea @stelcheck ?

lib/archivist/index.js
```diff
@@ -774,7 +774,13 @@ Archivist.prototype.set = function (topic, index, data, mediaType, encoding, exp
 	trackSize(topic, index, value);
 
 	value.set(mediaType, data, encoding);
-	value.touch(expirationTime);
+
+	if (!expirationTime) {
+		const config = configuration.getTopicConfig(topic);
+		value.touch(Date.now() / 1000 + config.ttl);
```
stelcheck: What will happen if config.ttl is not set?

Nokitoo: Right, I will modify it

Nokitoo: For add, do you mean touch ?

stelcheck: No, there should be an `archivist.add` normally.

Nokitoo: I don't see any archivist.add

stelcheck: Ok, might be some higher level stuff. Don't worry about it.

PR #234, opened 2018-04-23
details on vault setup errors, create migrates

### LINKED ISSUE #104: Cryptic error if you haven't run npm run archivist-create

Nokitoo:
Error log will hint the user that they need to run `archivist:create`
to setup up the storage.

`archivist:create` now implies `archivist:migrate`.

Fixes #104

### DIFF
--- a/lib/archivist/vaults/mysql/index.js
+++ b/lib/archivist/vaults/mysql/index.js
@@ -88,7 +88,13 @@ MysqlVault.prototype.open = function (cb) {
 
 	this.pool = this.mysql.createPool(this.config);
 
-	setImmediate(cb);
+	this.pool.getConnection((err) => {
+		if (err) {
+			this.logger.emergency('Failed to open mysql connection', this.config);
+		}
+
+		return cb(err);
+	});
 };
 
 MysqlVault.prototype.close = function () {
--- a/lib/tasks/archivist-create.js
+++ b/lib/tasks/archivist-create.js
@@ -1,14 +1,11 @@
 var async = require('async');
+var migrate = require('./migrate');
 
 
 exports.setup = function (mage, options, cb) {
 	async.series([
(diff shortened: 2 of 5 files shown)


lib/tasks/archivist-create.js
```diff
@@ -37,7 +34,7 @@ exports.start = function (mage, options, cb) {
 			return cb(error);
 		}
 
-		cb(null, { shutdown: true });
+		migrate.start(mage, options, cb);
```
stelcheck: I think this is meant to solve another issue reported in the tracker?

lib/archivist/vaults/mysql/index.js
```diff
@@ -88,7 +88,13 @@ MysqlVault.prototype.open = function (cb) {
 
 	this.pool = this.mysql.createPool(this.config);
 
-	setImmediate(cb);
+	this.pool.getConnection((err) => {
+		if (err) {
+			this.logger.emergency('Failed to open mysql connection', this.config);
```
stelcheck: Any reason to log for MySQL but nothing else?

Nokitoo: Because the MySQL error was not explicit

> MySQL error

```
[MAGE] Error during MAGE setup: getaddrinfo ENOTFOUND myhost myhost:3306
```

> Couchbase error

```
[MAGE] Error during MAGE setup: The remote host refused the connection. Is the service up?
  error: {
    "code": "connectionRefused",
    "type": "CouchbaseError"
  }
```

Indeed, we could log an error for each vault.

lib/tasks/archivist-create.js
stelcheck: Ah you are right. Thank you.

PR #237, opened 2018-04-25
Update ecmascript support

Nokitoo:
Supports only node >= v8.x.

This will allow us to use ES8 features in the engine such as async/await.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -7,6 +7,9 @@
     mocha: true
     node: true
     es6: true
+  parserOptions:
+    ecmaVersion: 8
+    sourceType: "module"
   rules:
     no-bitwise: 0
     camelcase:
--- a/docs-sources/includes/_Requirements.md
+++ b/docs-sources/includes/_Requirements.md
@@ -4,6 +4,10 @@
 
 <aside class="warning">
 MAGE only supports node from version 8.x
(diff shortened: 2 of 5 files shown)


.eslintrc.yml
```diff
@@ -7,6 +7,9 @@
     mocha: true
     node: true
     es6: true
+  parserOptions:
+    ecmaVersion: 8
```
stelcheck: Shouldn't 7 be sufficient?

docs-sources/includes/_Requirements.md
```diff
@@ -2,12 +2,8 @@
...
 ## Node.js
 
-<aside class="notice">
-MAGE always supports the "Active LTS" (Long Term Support) versions of Node.js.
-We recommend using the most recent Active LTS version, but we always try to support the Current Release as well.
-
-You can see what the LTS and Current versions are in the Node.js
-<a href="https://github.com/nodejs/Release#release-schedule">release schedule</a>.
+<aside class="warning">
+MAGE only supports node from version 8.x
```
stelcheck: Why delete the whole paragraph? 

Also, node -> Node.js

node: cli
Node.js: name of the framework

.eslintrc.yml
Nokitoo: No, otherwise you can't use async.

```javascript
// Unexpected token function
async function myFunction() {

}
```

PR #241, opened 2018-04-26
Fix mysql vault setup

Nokitoo:
Mysql pool was created in MysqlVault.open but it should be created in MysqlVault.setup as it's not doing any connection.
Also, MysqlVault.createDatabase and MysqlVault.dropDatabase need the pool when MysqlVault.open is not called.


Nokitoo:
We could call pool.getConnection to create one connection.

PR #243, opened 2018-04-27
Add config endpoint

### LINKED ISSUE #239: Modify how mage-sdk-js get Mage config
Currently for the mage-sdk-js, we need the client and server to be on the same repository. It's not very convenient and client/server should be separated.

Linked to the proposal in the issue #214, I would make this proposal:

1. **Automatically add a route on server-side to get the "client config"**

When starting the server, it would add a route to get the "client config" the same way we are doing for [httpBatchHandler](https://github.com/mage/mage/blob/master/lib/commandCenter/httpBatchHandler.js#L427).

2. **Modify/Rename `Mage.getClientConfig`**

As client and server should be separated, it may be a great idea to get a config without client `baseUrl`. We should do this in the sdk instead (Like in #214 proposal).
Thus, we could also rename it to `getConfig` or `getServerConfig`.

3. **In the mage-sdk-js, set the baseUrl and automatically get the config**

In the mage-sdk-js, we could just set an endpoint, and the sdk will take care of getting the server config and concatenate server config with `baseUrl`.

We could do something like this:
```javascript
const mageSdk = require('mage-sdk-js');
const endpoint = new URL(window.location.href);
endpoint.port = 1234;

mageSdk.setEndpoint(endpoint.href);

// Get config from server
mageSdk.init();

mageSdk.addModule('session', require('mage-sdk-js.session'));

mageSdk.setup(function(error) {
    // ...
})
 ```

Or: 

```javascript
const MageSdk = require('mage-sdk-js');
const endpoint = new URL(window.location.href);
endpoint.port = 1234;

// Create instance and get config from server
const mageSdk = new MageSdk(endpoint.href);

mageSdk.addModule('session', require('mage-sdk-js.session'));

mageSdk.setup(function(error) {
    // ...
})
 ```

I don't know how works [mage-sdk-unity](https://github.com/mage/mage-sdk-unity), but we could do the same. @stelcheck, how does mage-sdk-unity load Mage config ?

### PR DESCRIPTION
**Pull request by Nokitoo**
Expose /config endpoint to get the server config.

Fixes #239

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -479,3 +479,25 @@ exports.getConfigList = function () {
 exports.setLogger = function (loggerInstance) {
 	logger = loggerInstance.context('config');
 };
+
+exports.expose = function (mageInstance) {
+	const httpServer = mageInstance.core.httpServer;
+
+	function handler(req, res, path, query) {
+		if (!query.appName) {
+			res.setHeader();
+			res.writeHead(400, {
+				'content-length': '0'
+			});
+			res.end();
+		}
+
+		const config = mageInstance.getConfig(query.appName);
+		res.writeHead(200, {
+			'content-type': 'application/json'
+		});
+		res.end(JSON.stringify(config));
+	}
+
+	httpServer.addRoute(new RegExp('^/config'), handler, 'simple');
+};
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -479,25 +479,3 @@ exports.getConfigList = function () {
 exports.setLogger = function (loggerInstance) {
 	logger = loggerInstance.context('config');
 };
-
-exports.expose = function (mageInstance) {
-	const httpServer = mageInstance.core.httpServer;
-
-	function handler(req, res, path, query) {
-		if (!query.appName) {
-			res.setHeader();
-			res.writeHead(400, {
-				'content-length': '0'
-			});
-			res.end();
-		}
-
-		const config = mageInstance.getConfig(query.appName);
-		res.writeHead(200, {
-			'content-type': 'application/json'
-		});
-		res.end(JSON.stringify(config));
-	}
-
-	httpServer.addRoute(new RegExp('^/config'), handler, 'simple');
-};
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

File: lib/config/config.js [hunk 1]
```diff
@@ -479,3 +479,25 @@ exports.getConfigList = function () {
...
+		}
+
+		const config = mageInstance.getConfig(query.appName);
+		res.writeHead(200, {
+			'content-type': 'application/json'
+		});
+		res.end(JSON.stringify(config));
+	}
+
+	httpServer.addRoute(new RegExp('^/config'), handler, 'simple');
```
stelcheck: Wouldn't '/config' just work?

lib/config/config.js
```diff
@@ -479,3 +479,25 @@ exports.getConfigList = function () {
...
+				'content-length': '0'
+			});
+			res.end();
+		}
+
+		const config = mageInstance.getConfig(query.appName);
+		res.writeHead(200, {
+			'content-type': 'application/json'
+		});
+		res.end(JSON.stringify(config));
```
stelcheck: Why not cache the stringified value? Configuration will not change over a given runtime normally.

lib/config/config.js
stelcheck: Ah, I see. You get the app name.

stelcheck:
> getClientConfig has also been renamed to getConfig, 

This is the client config - we need a way to make a clear distinction between client config and server config.

lib/mage/Mage.js
```diff
@@ -817,23 +819,21 @@ Mage.prototype.isDevelopmentMode = function (feature) {
 };
 
 
-Mage.prototype.getClientConfig = function (appName, baseUrl) {
+Mage.prototype.getConfig = function (appName) {
```
stelcheck: We most likely need to still have the baseUrl to be injectable for dynamic environments (example: CI deployment)

lib/modules/config/index.js
```diff
@@ -0,0 +1,8 @@
+var mage = require('../../mage');
```
stelcheck: You should be able to require `lib/mage` if you want.

PR #247, opened 2018-05-09
Add options in typescript template tsconfig.json

Nokitoo:
Add experimentalDecorators and emitDecoratorMetadata options in typescript template tsconfig.json because in most of the cases mage-validator is used, and it needs theses options.


stelcheck:
I am not sure I am comfortable introducing experimental features in our template(s). Should mage-validator fall out of favour, we wouldn't be able to explain ourselves why we had things set up this way in the first place.

PR #248, opened 2018-05-10
Add built-in modules list

Nokitoo:
List all the built-in modules in the userguide.

### DIFF
--- a/docs-sources/includes/_Modules.md
+++ b/docs-sources/includes/_Modules.md
@@ -2,18 +2,20 @@
 
 ## External helper modules
 
-Some mage modules are optional but simplify the use of the library.
+External helpers modules are optional but simplify the use of the library.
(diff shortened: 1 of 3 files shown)


docs-sources/includes/_Modules.md
```diff
@@ -3,17 +3,19 @@
 ## External helper modules
 
 Some mage modules are optional but simplify the use of the library.
+Unlike other modules, they don't expose usercommands, they just act as helpers.
```
stelcheck: I think that is incorrect, e.g. archivist.

Nokitoo: Is there an archivist external module ?

stelcheck: ah sorry, I misread that.

Nokitoo: I will rephrase so there is no mistake

PR #249, opened 2018-05-11
Add more examples for manual requests in user guide

Nokitoo:
Add XMLHttpRequest and JQuery requests examples in user guide. The user may want to make requests to the server without using the client sdk. In that case, it could be difficult with the specific format of the requests.


stelcheck:
Wouldn't it make more sense to refer people to using mage-sdk-js instead?

Nokitoo:
There was a case where a user only used MAGE for outgame, and needed to make a call from ingame to store player score. Thus, using the client SDK for one call may not make a lot of sense.

stelcheck:
This is probably fine, but it's still not really what I think we should recommend in our user doc. There is generally more benefits to using the SDK than by maintaining your own lightweight SDK.

Alternatively, we could also add a "manual mode" in the SDK that skips the configuration steps for those cases.

Nokitoo:
We could add a method to send a single user command.
If mage/mage-sdk-js/pull/7 is merged we could do:
```javascript
mage.setEndpoint(
    'http://127.0.0.1:8080',
    'game'
);
mageSDK.sendCommand(
    'players'
    'register',
    {
        username: 'test'
        password: 'qdsqdsqds'
    }
);
```

Or, we could add a method which need no configuration at all:
```javascript
mageSDK.sendCommand(
    'http://127.0.0.1:8080/game/players.register',
    {
        username: 'test'
        password: 'qdsqdsqds'
    }
);
```

Which one do you think is more convenient ?

stelcheck:
```javascript
// Set endpoint but do not explicitly configure
mage.setEndpoint('http://127.0.0.1:8080', 'game', false);

mageSDK.sendCommand('players.register', {
    username: 'test'
    password: 'qdsqdsqds'
}, function (error, data) { /* ... */ });
```

Alternatively, `setEndpoint` could be set not to configure, and we could configure the first time we call `configure` intsead.

Nokitoo:
I created a PR mage/mage-sdk-js#8, we can close this PR.

PR #251, opened 2018-05-11
Remove archivist:migrate implicit call in archivist:create

Nokitoo:


stelcheck:
Why are we removing it? Please provide a description so that we may understand the purpose of this PR.

Nokitoo:
mage/mage/pull/234 introduced a bug.
archivist:create now implies archivist:migrate but archivist:create does not call mage.openVaults. Thus, if you have migrations it will fail as the database connection is not opened.

PR #255, opened 2018-05-16
Add msgStream documentation

### LINKED ISSUE #254: Document msgStream configuration.
Ref: https://github.com/mage/mage/blob/45b7de63797389bc4ed011ee105f7ec2396bcaae/lib/msgServer/config.yaml#L3-L5

Namely: 

  - transport preference
  - per transport configuration (e.g. heartbeat)

Nokitoo:
Explain
- how msgStream work and what is the link between msgStream and events
- how to configure msgStream

Fixes #254

### DIFF
--- a/docs-sources/includes/_Events.md
+++ b/docs-sources/includes/_Events.md
@@ -1,8 +1,51 @@
 # Events
 
-We have seen in the [States](#states) section of this user guide that
-states can be used to emit events between players. Let's dig a bit deeper
-into how this can be used.
+We have seen in the [States](#states) section of this user guide that states can be used to emit events between players.
+Before explaining how to send events, we will first see how MAGE send events to the clients, and how to configure the server to do so.
+
+## How does it work ?
+
+Events sending and receipt is done via **Message Stream**.
+
+Message Stream is a protocol used by MAGE servers and its clients in order to communicate. This protocol is implemented through different transports, which are the following:
+
+- **short-polling**: The client send a request to the server and get an **instant response** with the events. The client repeats this every X seconds to receive new events.
+- **long-polling**: The client send a request to the server and keep the connection open. The server will send a response with the events once one or multiple events destinated to the client are received. Then, the client open a new connection and repeat the process.
+- **websocket**: Real time events sending and receipt.
+
+If you need more information about Message Stream protocol, you can read this [documentation](https://github.com/mage/mage/tree/master/lib/msgServer/msgStream).
+
+In your MAGE config file, you can specify the priority of the transports which will be used by the client SDK to receive the events.
+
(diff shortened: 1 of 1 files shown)


docs-sources/includes/_Events.md
```diff
@@ -1,8 +1,56 @@
...
-states can be used to emit events between players. Let's dig a bit deeper
-into how this can be used.
+We have seen in the [States](#states) section of this user guide that states can be used to emit events between players.
+Before explaining how to send events, we will first see how MAGE send events to the clients, and how to configure the server to do so.
+
+## How does it work ?
+
+Events sending and receipt is done via **Message Stream**.
+
+**Message Stream** is a protocol used by MAGE servers and its clients in order to communicate. This protocol is implemented through different transports, which are the following:
```
stelcheck: I think we don't need to make **Message Stream** bold more than once - the first time its mentioned. Otherwise it feels a bit like we're screaming at people :)

docs-sources/includes/_Events.md
```diff
@@ -1,8 +1,56 @@
...
+> config/default.yaml
+
+```yaml
+server:
+    msgStream:
+        transports:
+            longpolling:
+                route: "/msgstream"
+                heartbeat: 60
+            shortpolling:
```
stelcheck: I think shortpolling also uses a heartbeat.

docs-sources/includes/_Events.md
```diff
@@ -1,8 +1,56 @@
...
+                route: "/msgstream"
```
stelcheck: We shouldn't encourage configuring the msgstream path... its pointless, and pretty much all integrations expect msgstream to be used anyway.

docs-sources/includes/_Events.md
Nokitoo: Ok, I didn't see this like that

docs-sources/includes/_Events.md
Nokitoo: shortpolling can use a heartbeat, but won't it become a longpolling then ?

docs-sources/includes/_Events.md
Nokitoo: What do you mean by all integrations ?

docs-sources/includes/_Events.md
stelcheck: heartbeat would refer to the frequency at which the client should poll I think... but might be an sdk config thing, not server... so forget what I said :)

docs-sources/includes/_Events.md
stelcheck: mage unity client, for instance

PR #260, opened 2018-06-18
Can ban user

Nokitoo:
Add two methods in auth module:

- `auth.ban`: add ban group acl (constant) to the user
- `auth.unban`: remove ban group acl (constant) from the user

Add unit tests testing the new behaviour.  The only untested path is in `CommandCenter.executeCommand`.

When a user is banned and tries to access a user command or login, it returns a `banned` error.

### DIFF
--- a/lib/commandCenter/commandCenter.js
+++ b/lib/commandCenter/commandCenter.js
@@ -540,6 +540,23 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
 		data: metaData
 	});
 
+	if (state.isBanned()) {
+		logger.info.data({
+			app: this.app.name,
+			module: cmdInfo.mod.name,
+			command: cmd.name,
+			user_acl: state.acl
+		});
+
+		return state.error('banned', null, function () {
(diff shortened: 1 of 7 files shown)


lib/commandCenter/commandCenter.js
```diff
@@ -540,6 +540,23 @@ CommandCenter.prototype.executeCommand = function (cmd, session, metaData, cb) {
...
+			module: cmdInfo.mod.name,
+			command: cmd.name,
+			user_acl: state.acl
+		});
+
+		return state.error('banned', null, function () {
+			// close the state and send the response
+
+			state.close(function (closeError, response) {
+				cb(null, response);
```
stelcheck: Shouldn't the closeError be handled?

Nokitoo: I used the same code we use for acl state `auth` error so I'm not sure.
But we could modify to `cb(closeError, response)`.

stelcheck: Its fine like that for now... let's see how it goes.

PR #271, opened 2019-02-20
Implement Redis list operation

lzubiaur:


lib/archivist/vaults/redis/index.js
```diff
@@ -92,6 +92,31 @@ RedisVault.prototype._unprefix = function (key) {
...
+
+	this.logger.error('scan:', key);
+
+	const rscan = (cursor) => {
+		this.client.scan([cursor, 'MATCH', key, 'COUNT', 100], function (error, data) {
+			if (error) {
+				return cb(error);
+			}
+
+			data[1].forEach((k) => set.add(k.toString()));
```
stelcheck: Why would we expect duplicates? If not, a simple array would do I think.

lzubiaur: yes, SCAN could return duplicates on a full iteration. that's a drawback of the command

PR #278, opened 2019-05-17
Consul improvements

Evertras:
* Instead of the PID, use the generated ID for the service so that a crash/restart will reuse the same key instead of creating dupes that will drop off and destroy our sanity as they bring down modules
* Increase rate of Consul session refreshes to half of TTL to ensure they remain alive even with latency spikes
* Don't lock on sessions so a replacement cluster can take over the key safely

### DIFF
--- a/lib/serviceDiscovery/engines/consul/index.js
+++ b/lib/serviceDiscovery/engines/consul/index.js
@@ -89,20 +89,19 @@ class ConsulService extends Service {
 	announce(port, metadata, cb) {
 		logger.debug(`Announcing service ${this.name} on port ${port}`);
 		this._announces.push({port, metadata});
-		this.session((err, id) => {
+		this.session((err) => {
 			if (err) {
 				return cb(err);
 			}
 
-			const key = [this.path, os.hostname(), process.pid].join('/');
+			const key = [this.path, os.hostname(), helpers.generateId(port)].join('/');
 			this._consul.kv.set({
 				key,
 				value: JSON.stringify({
 					port,
 					ip: this._ip,
 					data: metadata
-				}),
-				acquire: id
+				})
 			}, cb);
 		});
 	}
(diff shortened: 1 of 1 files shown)


lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -242,7 +241,7 @@ class ConsulService extends Service {
 				logger.error.data(err).log(`Failed to renew session ${this._sessionId}`);
 			}
 
-			this._keepaliveTimer = setTimeout(this._keepalive.bind(this, sessionId), this.ttl);
+			this._keepaliveTimer = setTimeout(this._keepalive.bind(this, sessionId), this.ttl / 2);
```
stelcheck: Shouldn't we change the default ttl value instead? See: https://github.com/mage/mage/blob/master/lib/serviceDiscovery/engines/consul/index.js#L16

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -89,20 +89,19 @@ class ConsulService extends Service {
...
 		logger.debug(`Announcing service ${this.name} on port ${port}`);
 		this._announces.push({port, metadata});
-		this.session((err, id) => {
+		this.session((err) => {
 			if (err) {
 				return cb(err);
 			}
 
-			const key = [this.path, os.hostname(), process.pid].join('/');
+			const key = [this.path, os.hostname(), helpers.generateId(port)].join('/');
```
stelcheck: Should we remove the hostname? In dockerized environment, the hostname will be the container ID, and a new container will have a different ID but might have the same generated ID.

Evertras: That's a good point.  I'll switch it to `this.path, helpers.generateId` instead.

lib/serviceDiscovery/engines/consul/index.js
Evertras: The default TTL doesn't matter so much, we want to make sure that we're actually hitting Consul well before the TTL would expire.  Doing it at the TTL interval means if any network hiccup hits we're at risk of the service 'going down' and this whole issue rearing up again.

Specifically, from Consul docs:

> When a TTL is being used, clients should be aware of clock skew issues: namely, time may not progress at the same rate on the client as on the Consul servers. It is best to set conservative TTL values and to renew in advance of the TTL to account for network delay and time skew.

stelcheck: OK - let's try it with half-time then.

PR #280, opened 2019-05-20
Use Consul locks to ensure expired services get removed

Evertras:
The earlier PR had a hidden bug: the value wasn't being expired properly.  This reintroduces consul locks to fix that issue and automatically takes over the expired key in the event of a fast restart by destroying the old session.

### DIFF
--- a/lib/serviceDiscovery/engines/consul/index.js
+++ b/lib/serviceDiscovery/engines/consul/index.js
@@ -5,6 +5,7 @@ const requirePeer = require('codependency').get('mage');
 const consul = requirePeer('consul');
 
 const mage = require('lib/mage');
+const MageError = require('../../../mage/MageError');
 const Service = require('../../service').Service;
 const ServiceNode = require('../../node').ServiceNode;
 const helpers = require('../../helpers');
@@ -113,7 +114,10 @@ class ConsulService extends Service {
 					// If we can contact Consul but we can't set the key, there is no 'err'; the
 					// returned result is simply set to false so we need to check for it here
 					if (result === false) {
-						return cb(new Error('Could not set Consul KV for service (likely due to another lock)'));
+						return cb(new MageError({
+							message: 'Could not set Consul KV for service',
+							details: 'Consul was contacted successfully, but the key was unable to be set after deleting existing sessions.  If this persists, check if the Consul KV store is configured correctly.'
+						}));
 					}
 
 					cb(null);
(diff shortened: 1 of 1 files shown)


lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -88,20 +89,52 @@ class ConsulService extends Service {
...
+					acquire: id
+				}, (err, result) => {
+					if (err) {
+						return cb(err);
+					}
+
+					// If we can contact Consul but we can't set the key, there is no 'err'; the
+					// returned result is simply set to false so we need to check for it here
+					if (result === false) {
+						return cb(new Error('Could not set Consul KV for service (likely due to another lock)'));
```
stelcheck: Could you look into using MageError, which would provide a better logging facility built-in?

lib/serviceDiscovery/engines/consul/index.js
```diff
@@ -60,7 +60,8 @@ class ConsulService extends Service {
 
 		this._consul.session.create({
 			behavior: 'delete',
-			ttl: `${Math.ceil(this.ttl / 1000)}s`
+			ttl: `${Math.ceil(this.ttl / 1000)}s`,
+			lockdelay: '0s'
```
stelcheck: Should this be made configurable instead?

stelcheck (changes requested):

lib/serviceDiscovery/engines/consul/index.js
Evertras: Lockdelay needs to be 0 in order for the rest to work.  If the lock delay is anything but zero, deleting the session doesn't actually release the lock until the delay has passed.  The intent from Consul's end is to provide lock holders with the opportunity to see their session be invalidated and have a window to fix it, but we're not doing that.

lib/serviceDiscovery/engines/consul/index.js
Evertras: Gotcha, I'll change it.

Evertras: After looking into this, I'm not sure that's the correct way to go.  MageError seems geared towards sending clients error information, but this is a critical error that will prevent the application from even starting up as it's an uncaught exception down the pipe and regular Error is being used elsewhere in the file.  Thoughts?

stelcheck: IIRC it will help you provide more contextual data once the error gets logged. So I believe it should provide additional value, but you might want to test to see how that would look.

Evertras: It does give the description field which I can put some stuff in.  The code/level is wasted but it's at least a bit of extra info, yeah.  I'll use it.  Pushing shortly.

stelcheck (approved):

stelcheck (approved):

